PR #46, opened 2019-09-09
(WIP) Documentation in Github

atschalz:
This is a proposal for the documentation. Please have a look at it and give me feedback about the general layout. Is this fine or should the focus be on something else?

For the following points concrete feedback is needed:
- I did not work with other commands than start (and a little resume), so I copied the help output for the moment. Is this enough or should there be more information in terms of the output of the command or something else?

- I did not find all the relevant citations for the models and I am unsure about the correct citations for the datasets. Please check if the citations are correct and tell me which paper is needed for the missing citations.

Further, a short description about training types and evaluation methods will be added. Please let me know if something else important is missing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,7 @@
 # kge
 
-# Run
-- See available commands: `kge.py --help`
-
-# Installation
-- locally: run `pip install -e .`
+# Todo
+- Include links to models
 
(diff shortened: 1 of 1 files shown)


rgemulla:
Outdated.

PR #55, opened 2019-10-28
(Finished) Triple Classification

atschalz:
The implementation is finished and the evaluation results were compared to previous results to approve the implementation. An example file "examples/toy-complex-train-tripleclass.yaml" is added which trains a model and validates it on triple classification. 

Most important user information considering the configurations:
- eval.type has to be set to triple_classification to evaluate on the triple classification task
- valid.metric has to be set to accuracy
- if metrics_per.relation is set to True, the accuracy and precision per relation are also computed. This is useful since the results vary across different relations. Such a comparison is also frequently done in the literature.
- In triple classification, every triple in the validation and test data is corrupted to create false triples. This is always done when running TC. Naturally in TC, the results vary due to the corrupted triples. If eval.triple_classification_random_seed is set to True, it is guaranteed, that the same data is generated in the corruption. This makes comparisons across models possible. Nevertheless, we cannot access if we have generated a good or bad corrupted dataset, since this would be time-consuming.

Further information considering the implementation:
- The main change to the framework can be found in triple_classification.py. Furthermore, a new sampling function explicitly for that task is implemented. If it becomes necessary in the future, other sampling function could be used as well for the TC task. 
- To find the thresholds, another method than originally used by Socher et al. 2013 is used. The original implementation is also provided, but it is very inefficient. You can have a look at it and decide if it is worth to keep, I would just delete it and use the faster implementation. More detailed information can be found in the code documentation (triple_classification.py, lines 119-194), If this is not enough, please ask and I will explain in more detail what was done.
- For calculating the final accuracy, an sklearn function is used. This could also be changed.
- For generating corrupted triples, random.choice is used. I did not find a PyTorch equivalent that fits, if necessary this probably can also be replaced by another function.
- To retrieve the scores, model.score_spo() is used in triple_classification.py, line 62/68. Unfortunately, this doesn't work for RESCAL, since it 'cannot handle combine="spo" '. If necessary, the retrieval of the scores can be changed, but it would be better to make this function work also for RESCAL (if possible).

If there are any questions, feel free to ask. I am looking forward for further feedback!


rgemulla:
Closed for now since required some effort to rewrite to current more stable API.

PR #56, opened 2019-10-29
(Finished) Readable top k predictions per triple as trace, log and terminal output

atschalz:
This contribution makes it possible to manually inspect the predictions a model makes for test triples. For every test triple, the k predictions with the highest scores are added to the trace and if possible added in an interpretable readable form to the log file and outputted.

User guidelines for the configurations:
- For the predictions to be readable, a file "entity_map_realnames.txt" has to be added to the data folder. If such a file does not exist, only the entity indices are reported. The file should contain a mapping from metadata string IDs to real entity names. 
E.g.:  "/m/027rn	"Dominican Republic"@e
For the beginning one file for freebase mappings is added to the toy data folder. To use this contribution for real data, mappings to real entity names for all datasets have to be found (preferably english only and as complete as possible).

- If eval.top_k_predictions.predict is set to True, the best k predictions are added to the trace. If a mapping to real world entity names is available, the predictions are also added to the log file with the real entity and prediction names.

- eval.top_k_predictions.k specifies the no. of predictions to be inspected

- If eval.top_k_predictions.print is set to True, the real name predictions are outputted to the terminal. This specification is added, because for large datasets, we can have a lot of triples to make predictions for. This is also the reason why by default everything is set to False.

Details considering the implementation:
- A dictionary called entities_map containing the required mapping is added to the dataset object. For loading it, a new function was created, which is very similar to the already existing _load_map function. The already existing function converts the key of the map to an integer, which cannot be done for the real names mapping (string). Changing the function would lead to a less readable function and would also change the function for the already existing mappings, so a new function was created for keeping this separated. 

- If a mapping is not present or named wrong, this is printed to the terminal and the dictionary in the dataset object is set to "Not available". In this case the trace and log file only contain the entity indices.

- The main part of contribution is the function get_top_k_predictions in eval.py, which can be used as a post_epoch_trace_hook by every evaluation type (entity_ranking, triple_classification, etc.). The function consists of the following main parts:

1. The scores for all test triples are retrieved. This is also done in the entity_ranking script, which means that the operation is executed twice in an evaluation task. To maintain the flexibility to use the function for other evaluation types which may use differently retrieved scores (e.g. triple_classification), the decision was to better do this again in the function.

2. Find the best predictions for every test triple. We create a dictionary, containing entries as {test triple: [1st predicted triple,  ..., kth predicted triple]}. Here we only operate with the entity indices.

3. The entries of this dictionary are added to the trace. Usually, it would be better to also add the real names to the trace if possible, but they often contain special characters, so that while writing out the trace, the strings get corrupted, which leads to an unreadable output. At the moment I don't see a way to fix this.

4. If a mapping to real entity names is available, the entity indexes as well as the relation indexes are replaced with the real names (stored in a new dictionary). If an entity name is not present in the real names mapping, the string metadata key is used instead. This new dictionary with the interpretable triple predictions is added to the log file in a readable way. If specified, it is also printed in the terminal.

As has been said, this contribution becomes more useful if there are good mappings to entity names. Thanks in advance for any kind of feedback!

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -205,6 +205,9 @@ eval:
   num_workers: 0
   pin_memory: False
   trace_level: example            # example or batch or epoch
+  predict_output: True
+  predict_output_k: 5
+  predict_output_print: True
 
   # Metrics are always computed over the entire evaluation data. Optionally,
   # certain more specific metric can be computed in addition.
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -200,14 +200,16 @@ valid:
 eval:
   data: valid
   type: entity_ranking
-  hits_at_k_s: [1, 3, 10, 50, 100, 200, 300, 400, 500, 1000]     # ks for HITS@k
+
+  predict_output: True
+  predict_output_k: 5
+  predict_output_print: True
+  hits_at_k_s: [1, 3, 10, 50, 100, 200, 300, 400, 500, 1000]     # k for HITS@k
   batch_size: 100
   num_workers: 0
   pin_memory: False
   trace_level: example            # example or batch or epoch
-  predict_output: True
-  predict_output_k: 5
-  predict_output_print: True
+
 
   # Metrics are always computed over the entire evaluation data. Optionally,
   # certain more specific metric can be computed in addition.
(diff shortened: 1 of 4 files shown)


rgemulla:
Closed for now since required some effort to rewrite to current more stable API.

PR #64, opened 2019-12-14
frequency based negative sampling

AdrianKs:
This is a basic frequency based negative sampling approach.


rgemulla:
What's the state of this PR (it's very useful!)?

AdrianKs:
It needs to be updated to the new structure of the sampler.
I will update it and let you know.

AdrianKs:
The code is ready for review.

kge/config-default.yaml
```diff
@@ -188,7 +188,8 @@ KvsAll:
 # Options for negative sampling training (train.type=="negative_sampling")
 negative_sampling:
   # Type of sampler
-  sampling_type: uniform
+  sampling_type: uniform    # uniform, frequency
```
rgemulla: Add a doc string that explains the sampler types.

kge/config-default.yaml
```diff
@@ -188,7 +188,8 @@ KvsAll:
...
+  symmetric_prior: 1        # this prior is used to smooth counts in frequency sampling
```
rgemulla: Rename to ``frequency.smoothing`` and add a doc string.

kge/util/sampler.py
```diff
@@ -240,3 +241,51 @@ def _filter_and_resample_numba(
...
+    Can be smoothed with a symmetric prior.
+    """
+    def __init__(self, config, configuration_key, dataset):
+        super().__init__(config, configuration_key, dataset)
+        self.samplers = []
+        alpha = self.get_option("symmetric_prior")
+        for slot in SLOTS:
+            vocab = np.arange(self.vocabulary_size[slot].item())
+            unique, counts = np.unique(
+                dataset._triples["train"][:, slot], return_counts=True
```
rgemulla: Use ``dataset.train()``

rgemulla: You can get the counts directly in the format you need using ``np.bincount(y, minlength=<vocab_size>)``.

rgemulla: Then simply add the smoothing value.

kge/util/sampler.py
```diff
@@ -240,3 +241,51 @@ def _filter_and_resample_numba(
...
+            )
+
+            # smooth counts with the symmetric prior alpha
+            zero_counts = vocab[~np.isin(vocab, unique)]
+            sort_index = np.argsort(np.concatenate([unique, zero_counts]))
+            smoothed_counts = (
+                np.concatenate([counts, np.zeros(len(zero_counts))])[sort_index] + alpha
+            )
+
+            self.samplers.append(
```
rgemulla: self._multinomials

rgemulla:
Is it easy to also add the "fast" implementation? (Could also be a separate PR later on.) From my point of view, it may suffice to take this method from KgeUniformSampler and change the line starting with ``new_samples=`` accordingly.

AdrianKs:
I addressed your comments.

I think the "fast" implementation is still some work, since in the new_samples= line we directly call the numpy random function.
For frequency sampling we would call the torch multinomial function, which probably is not supported by numba.
When we call the python sample method from the fast numba method it should become slow again.

rgemulla:
Thanks!

PR #66, opened 2020-01-17
(WIP) Rescal memory fix

AdrianKs:
this is a first version of the rescal memory fix.
Here we calculate the sp and po scores for each unique relation.

Unfortunately the current implementation is still way too slow


rgemulla:
Outdated & addressed otherwise.

PR #70, opened 2020-02-14
Faster negative sampling

Nzteb:
Reviews from the original PR have been addressed.

Approximate timings (as from the orig. PR)

30 samples; avg.  seconds per epoch; fb15k

filter type|1 worker|4 workers
-|-|-
no filter|16|14
fast|22|19
standard|42|19

60 samples; avg.  seconds per epoch; fb15k

filter type|1 worker|4 workers
-|-|-
no filter|19|20
fast|25|22
standard|47|21

120 samples; avg.  seconds per epoch; fb15k

filter type|1 worker|4 workers
-|-|-
no filter|29|29
fast|33|31.5
standard|61|31

### DIFF
--- a/kge/util/sampler.py
+++ b/kge/util/sampler.py
@@ -1,4 +1,5 @@
 from kge import Config, Configurable, Dataset
+from kge.indexing import index_where_in
 import torch
 from typing import Optional
 import numpy as np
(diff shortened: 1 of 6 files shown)


kge/util/sampler.py
```diff
@@ -72,11 +90,25 @@ def sample(
         else:
             negative_samples = self._sample(positive_triples, slot, num_samples)
         if self.filter_positives[slot]:
-            if self.shared:
-                raise ValueError(
-                    "Filtering is not supported when shared negative sampling is enabled."
```
Nzteb: I have moved this to the constructor.

rgemulla:
Thanks!

PR #71, opened 2020-02-20
(WIP) Sp po unique

AdrianKs:
This is a memory friendly and faster negative sampling implementation.
Instead of scoring against all entities in sp_po we only score against unique(batch, negatives).

image

sp_po_unique_all (should maybe be called sp_po_unique_tuple) is an extension to sp_po_unique and only scores every unique tuple against the needed unique entities. Gives small improvements in memory, especially with RESCAL.

image

Do we want to replace sp_po with sp_po_unique or keep both?


rgemulla:
The semantics for sp_po and sp_po unique is exactly the same right (each unique target entity is embedded precisely once)? In this case, I think we do not need sp_po anymore; I fail to see any reason to use it.

About sp_po_all: My understanding is that you mean "unique prefix", not unique tuple, right? The method is then beneficial only if a batch contains repeated prefixes. This may be unlikely when batches are formed randomly, but for other forms of batch construction (which we do not do right now) it may actually help a lot.

As for naming, I suggest to call one method sp_po (the one currently called sp_po_unique) and the other sp_po_unique (the one currently called sp_po_unique_all).

Is there some non-negligible overhead associated with sp_po_unique_all? If not, we could just always use this. More generally, how do these methods affect runtime?

rgemulla:
Perhaps the old method could me renamed to sp_po_naive, in case we want to keep it.

samuelbroscheit:
The semantics are not the same. This has the same issue with dropout as the other method that was using "unique" on the batch's indexes. 

However, we did also agree on making the semantics for the whole framework like this.

AdrianKs:
spo and sp_po have a different semantic in terms of dropout, but for sp_po and sp_po_unique the semantics are exactly the same.
The semantic changes again for sp_po_unique_all

I will compare epoch times of these methods again with a smaller setting, where we could actually see an overhead

AdrianKs:
so on a day to day setting on the dataset fb15k-237 we can obviously see an overhead, since we usually use every entity in one batch:
complex
bs 1000

|  Implementation  | Negative samples | Shared | Epoch Time |
|:----------------:|------------------|--------|------------|
| sp_po            |             1000 | False  |      14.4s |
| sp_po_unique     |             1000 | False  |      16.3s |
| sp_po_unque_all  |             1000 | False  |      15.6s |
| sp_po            |              100 | False  |       9.8s |
| sp_po_unique     |              100 | False  |      10.9s |
| sp_po_unique_all |              100 | False  |      10.2s |
| sp_po            |             1000 | True   |      11.0s |
| sp_po_unique     |             1000 | True   |       4.8s |
| sp_po_unique_all |             1000 | True   |       4.8s |
| sp_po            |              100 | True   |       9.8s |
| sp_po_unique     |              100 | True   |       2.2s |
| sp_po_unique_all |              100 | True   |       2.1s |



but as soon as num_unique_entities_in_batch << num_unique_entities, we see improvements
yago3-10
complex
batch size 1000

|  Implementation | Negative samples | Shared | Epoch Time |
|:---------------:|------------------|--------|------------|
| sp_po           |               50 | False  |     299.0s |
| sp_po_unique    |               50 | False  |     119.5s |
| sp_po_unque_all |               50 | False  |     113.6s |



sp_po_unique_all gives timeimprovements over sp_po_unique, since both yago and 237 have a small number of relaions, I think

rgemulla:
Thanks! I think it's best to stick to sp_po_naive and sp_po_unique, esp. since they have the same semantics. The improvements from unique_all appear to be rather marginal.

rgemulla:
BTW: where does the overhead of sp_po_unique come from? Is it the duplicate elimination?

AdrianKs:
The overhead is due to the fact, that we usually use all entities in small sets like 237 with larger batch sizes and num_samples.
Therefore we call torch.unique(set_containing_all_entities). This sorts all the entities and eliminates duplicates. Additionally we create the inverse indexes for the unique entities to the original tensor.

So it results in an overhead for eliminating and sorting, but does not reduce scoring in settings like this.

i will rename the functionalities and remove sp_po_unique_all

AdrianKs:
All done. The code is ready for review.

These changes also fix a little bug with the negative sampling auto functionality.

rgemulla:
Github currently does not seem to work, so first comments here:

1. The documentation in config_default.yaml needs to be updated. I suggest to use a list with bullet points, one per type of implementation. Also explain what the strategies do in one sentence.

2. unique_triples is a misnomer. Rename to sth. like unique_indexes or unique_targets or so.

3. Perhaps add 2-3 comments to sp_po implementation to describe it

4. Add slot == P; this shoulbe be easy since we now have score_so. Or is there any complication?

rgemulla:
5. Add an entry to CHANGELOG.md

AdrianKs:
I addressed the points and renamed the negative sampling implementations.

kge/config.py
```diff
@@ -619,6 +621,16 @@ def rename_value_re(key_regex, old_value, new_value):
...
+    # 26.02.2020
+    rename_key(
+        "negative_sampling.implementation.spo",
+        "negative_sampling.implementation.triple",
+    )
+    rename_key(
+        "negative_sampling.implementation.sp_po",
+        "negative_sampling.implementation.batch",
+    )
+
```
rgemulla: This does not work. You need to rename the value for key negative_sampling.implementation. Please adapt and test.

rgemulla: Everything else looks great.

AdrianKs: fixed

rgemulla:
Thanks!

PR #77, opened 2020-03-04
Add vanity CLI entrypoint

### LINKED ISSUE #76: Use entry_points to make vanity CLI script
Rather than providing kge.py, you can use the `console_scripts` entry point to define a CLI that automatically gets installed and made available in the shell with `pip install .`

Would be happy to send a PR!

cthoyt:
Closes #76

- The encapsulation of the `if __name__ == '__main__'` makes it possible to import the code that runs the CLI from anywhere
- The move from `kge.py` inside the package makes it possible to run it after install with `python -m kge.cli`
- The addition of `__main__.py` makes it possible to run the CLI from `python -m kge` since it's imported
- The addition of the `entry_points` keyword in `setup.py` makes the installer add a CLI script called `kge` that automatically runs the main function in `cli.py`.

I'd be happy to answer any questions


cthoyt:
Not sure why there's all the config files in the diff.. I'm not very skilled with Git so maybe this happened because there were some edits to the repo after I forked it

rgemulla:
Great, cherry-picked & thanks!

cthoyt:
You’re very welcome! Happy embedding and greetings from Bonn :)

PR #80, opened 2020-03-06
add weighted regularization to wikidata parameters in readme

AdrianKs:
This is probably an important information, since non-weighted does not work with a dataset of this size.


rgemulla:
Thanks!

In the future, we should probably implement non-weighted regularization differently then. In fact, whether we regularize the batch entities or all entities, and whether regularization is unweighted or weighted can be done independently. I'll open an issue to not forget this.

PR #83, opened 2020-03-11
Updated dump trace command documentation

Nzteb:


kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
     )
-
+    parser_dump_trace.add_argument(
+        "--train",
+        action="store_const",
+        const=True,
+        default=False,
+        help=(
+            "Include entries from training jobs (Note: if none of --train --valid"
+            " --test is set, this will activate all of them jointly)."
```
rgemulla: Confusing: write in parens "(enabled when none of --train, --valid, or --test is specified)"

kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
+            " if none of --train --valid --test is set, this will activate"
+            " all of them jointly)."
+        ),
+    )
+    parser_dump_trace.add_argument(
+        "--search",
+        action="store_const",
+        const=True,
+        default=False,
+        help="Dump the tracefile from a search job. If a tracefile from search jobs is"
```
rgemulla: *of* a search job

kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
+        "--keys",
+        "-k",
+        nargs="*",
+        type=str,
+        help=(
+            "A list of 'key' entries (separated by space). For every entry,"
+            " the corresponding values of 'key' will be searched in the config and"
+            " trace entries and in the csv file a column with name 'key' will be added."
+            " The same special keys as for --keysfile can be used but must be"
+            " surrounded by quotations ''. "
```
rgemulla: This is unlcear. You should say that each KEY is form "KEY_NAME" or "NEW_KEY_NAME=KEY_NAME" and then just refer to the description above

kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
     )
-
     parser_dump_trace.add_argument(
         "--max_epoch",
         default=False,
         help=(
-            "Specifies the epoch in the trace"
-            "from where to start processing backwards"
+            "Specifies the max epoch number in the tracefile"
+            " from where to start processing backwards."
```
rgemulla: Unclear. How does this differ from truncate?

kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
-        "--train",
-        "--valid",
-        "--test",
-        "--search",
+    parser_dump_trace.add_argument(
         "--yaml",
+        action="store_const",
+        const=True,
+        default=False,
+        help="Dump the full tracefile to stdout.",
```
rgemulla: Not true. Just means do not use CSV, but full trace+config entries. Which entries depends on the other options.

kge/util/dump.py
```diff
@@ -113,83 +113,167 @@ def _add_dump_trace_parser(subparsers_dump):
...
+        const=True,
+        default=False,
+        help="Exclude default keys from the csv file.",
+    )
+    parser_dump_trace.add_argument(
+        "--timeit",
+        action="store_const",
+        const=True,
+        default=False,
+        help="Output timing information of the command.",
```
rgemulla: Can't we drop this completely?

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
     parser_dump_trace = subparsers_dump.add_parser(
         "trace",
         help=(
-            "Process and dump trace to stdout and/or csv. The trace will be processed "
-            "backwards, starting with a specified job_id."
+            "Dump trace to csv (default) and/or stdout. The tracefile will be processed"
```
rgemulla: It's always stdout. So it should read "to stdout as CSV (default) or YAML"

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
...
+            " or --test is specified)."
+        ),
+    )
+    parser_dump_trace.add_argument(
+        "--valid",
+        action="store_const",
+        const=True,
+        default=False,
+        help=(
+            "Include entries from validation during training and evaluation on"
```
rgemulla: From validation or evaluation jobs on the valid split

kge/util/dump.py
Nzteb: Truncate only works together with a checkpoint. When a checkpoint is provided then --truncate will enforce the max_epoch number will be the one the checkpoint holds. On the other hand, --max_epoch works in any case and just sets the max_epoch number. Using both together does not make sense and I added a check to enforce this.

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
...
+        help="Dump the tracefile of a search job. The options --train, --valid, and"
+             " --test are not applicable.",
+    )
+    parser_dump_trace.add_argument(
+        "--keysfile",
+        default=False,
+        help=(
+            "A path to a file which contains lines in the format"
+            " 'new_key_name'='key_name'. For every line in the keys file, the command"
+            " will look for the value of 'key_name' in the trace entries and config and"
```
rgemulla: trace entries (first) and config (second)

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
...
     )
-
     parser_dump_trace.add_argument(
         "--max_epoch",
         default=False,
         help=(
-            "Specifies the epoch in the trace"
-            "from where to start processing backwards"
+            "Specifies the max epoch number in the tracefile"
+            " from where to start processing backwards. Can not be used with --truncate"
```
rgemulla: Cannot. What's the point of this option?

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
...
         action="store_const",
         const=True,
         help=(
-            "If a checkpoint is used (by providing one explicitly as source or by "
-            "using --checkpoint), --truncate will define the max_epoch to process as"
-            "specified by the checkpoint"
+            "When a checkpoint is used (by providing one explicitly as source or by"
+            " using --checkpoint), --truncate will define the max_epoch number to"
+            " process as provided by the checkpoint. Can only be used when a checkpoint"
+            " is provided and it cannot be used together with --max_epoch."
```
rgemulla: Likewise.

kge/util/dump.py
```diff
@@ -113,90 +113,174 @@ def _add_dump_trace_parser(subparsers_dump):
...
-        "--valid",
-        "--test",
-        "--search",
+    parser_dump_trace.add_argument(
         "--yaml",
+        action="store_const",
+        const=True,
+        default=False,
+        help="Instead of using a CSV file with a subset of entries, dump the full"
+             " trace file. Additional entries from the config can be added with the"
```
rgemulla: Still not true.

kge/util/dump.py
```diff
@@ -214,15 +298,16 @@ def _dump_trace(args):
             checkpoint_path = Config.get_best_or_last_checkpoint(args.source)
         folder_path = args.source
         if not args.checkpoint and args.truncate:
-            raise ValueError(
-                "You can only use --truncate when a checkpoint is specified."
-                "Consider using --checkpoint or provide a checkpoint file as source"
+            print(
+                "--truncate can only be used when a checkpoint is specified."
```
rgemulla: Add space after .

kge/util/dump.py
Nzteb: It also includes validation entries when one validates on the training data. That's why I wrote it like that.

kge/util/dump.py
Nzteb: I think you mean it the other way around?

kge/util/dump.py
Nzteb: I think it makes sense. It allows the user to just dump, say, entries of epochs 1-100 independent of a checkpoint or anything else. Apart from that, you asked me to add this option.

kge/util/dump.py
Nzteb: Likewise. We discussed we want as default when a checkpoint is provided that sill all epochs of the job from that checkpoint would be included. But to also have to flexibility to only process up until the epoch which is provided from the checkpoint which is what --truncate does.

kge/util/dump.py
Nzteb: Please make a suggestion.

kge/util/dump.py
Nzteb: Thanks

kge/util/dump.py
rgemulla: Yes, but it's redundant. Compare "evaluation on the valid split" and "validation and evaluation on the valid split"

kge/util/dump.py
rgemulla: Trace is tried first, right (it should be)

kge/util/dump.py
rgemulla: Why not call this option --truncate_epoch

kge/util/dump.py
rgemulla: And this one just --truncate

kge/util/dump.py
rgemulla: It's not the full trace file, but the same subset that would have been used for the CSV. So simply: "Dump YAML instead of CSV"?

kge/util/dump.py
Nzteb: I find --max_epoch (which was your suggestion) more intuitive than --truncate_epoch. But I can change it.

kge/util/dump.py
Nzteb: It is called just --truncate already. What do you mean?

kge/util/dump.py
rgemulla: How about using only --truncate with an optional numeric argument. If given, then that epoch is used. If not given, then the checkpoint epoch is used (if a checkpoint is specified; otherwise error).

kge/util/dump.py
rgemulla: Right.

:)

kge/util/dump.py
Nzteb: I'm not sure if I understand you correctly. The --csv option only dumps the default keys whereas the --yaml option directly dumps the complete  trace entry. It does not filter out anything, which is how I was instructed. So what do you mean with the "same subset"

kge/util/dump.py
Nzteb: For the keys from --keysfile and --keys, at the moment, the config is searched and then the trace entry is searched. So then this has to be changed?

kge/util/dump.py
rgemulla: Yes, the complete trace entry for the subset of selected trace entries. Not all trace entries in the trace file.

kge/util/dump.py
rgemulla: Yes, that's unintenional; please change. It probably only affects a small number of keys though.

kge/util/dump.py
Nzteb: Ok, we mean the same. Complete trace is indeed misleading.

kge/util/dump.py
Nzteb: But --truncate can only be used together with a checkpoint. --max_epoch is independent of checkpoints.

kge/util/dump.py
Nzteb: Okey I will change that.

kge/util/dump.py
rgemulla: I suggest to change that.

--truncate                   -> use epoch of checkpoint (required)
--truncate 50              -> truncate at epoch 50 (no matter whether or not a checkpoint is specified)

--max-epochs is then not needed anymore.

Nzteb: If you want this I can do it (As long as the parser allows for arguments of this kind?).

rgemulla: Yes, please go ahead. It's the cleanest and most intuitive way to do this.

rgemulla:
Thanks a lot!

PR #86, opened 2020-03-24
Support (s,?,o) queries for KvsAll training

vonVogelstein:


vonVogelstein:
As discussed with Rainer, added the option "use_os" for KvsAll training. The option allows to train not only on the rows and columns created from the sp and po pairs, but also on the tubes created from the so pairs

kge/config-default.yaml
```diff
@@ -220,6 +220,9 @@ KvsAll:
   # ( https://github.com/TimDettmers/ConvE/blob/853d7010a4b569c3d24c9e6eeaf9dc3535d78832/main.py#L156) with a default value of 0.1.
   label_smoothing: 0.0
 
+  # If set, training also runs on subject object (so) pairs additionally to sp and op pairs
```
rgemulla: Please rebase to master and then change options to: 
```
queries:
   sp_: True
   s_o: False
   _po: True
```
I.e., each of the KvsAll prediction tasks can be enabled/disabled individually.

kge/job/train.py
```diff
@@ -583,55 +596,65 @@ def collate(batch):
...
-                is_sp[batch_index] = 1 if example_index < num_sp else 0
-                if is_sp[batch_index]:
+                key_type[batch_index] = 1 if example_index < num_sp else (0 if example_index < end_po else 2)
+                if key_type[batch_index] == 1:
                     keys = self.train_sp_keys
                     offsets = self.train_sp_offsets
                     values = self.train_sp_values
-                    sp_po_col_1, sp_po_col_2, o_s_col = S, P, O
-                else:
+                    sp_po_so_col_1, sp_po_so_col_2, o_s_p_col = S, P, O
```
rgemulla: try to use more intuitive variable names: query_col1, query_col2, target_col for example

kge/job/train.py
```diff
@@ -583,55 +596,65 @@ def collate(batch):
...
                 ]
                 current_index += size
 
             # all done
             return {
-                "sp_po_batch": sp_po_batch,
+                "sp_po_so_batch": sp_po_so_batch,
                 "label_coords": label_coords,
-                "is_sp": is_sp,
+                "key_type": key_type,
```
rgemulla: query_type

kge/job/train.py
```diff
@@ -672,15 +706,28 @@ def _process_batch(self, batch_index, batch) -> TrainingJob._ProcessBatchResult:
...
             )
-            loss_value_po = self.loss(scores_po, labels[po_indexes,]) / batch_size
+            loss_value_po = self.loss(scores_po, labels_po) / batch_size
             loss_value += loss_value_po.item()
             forward_time += time.time()
             backward_time = -time.time()
             loss_value_po.backward()
             backward_time += time.time()
 
+        # forward/backward pass (so)
```
rgemulla: That's lots of repeated code. Why not loop over ```["sp_", "s_o", "_po"]``` and use an if-elif-else for computing scores/loss?

vonVogelstein:
Adjusted variable names, added feature request (allow all combinations of query types) and adjusted all code to work on dictionaries over query types instead of individual variables per query type

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
+                        example_index -= prev_end
+                        num_ones += self.offsets[query][example_index + 1]
+                        num_ones -= self.offsets[query][example_index]
+                        break
+                    prev_end = end
 
             # now create the results
-            sp_po_batch = torch.zeros([len(batch), 2], dtype=torch.long)
-            is_sp = torch.zeros([len(batch)], dtype=torch.long)
+            key_batch = torch.zeros([len(batch), 2], dtype=torch.long)
```
rgemulla, line 598: -> queries

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
-                else:
-                    example_index -= num_sp
-                    keys = self.train_po_keys
-                    offsets = self.train_po_offsets
-                    values = self.train_po_values
-                    o_s_col, sp_po_col_1, sp_po_col_2 = S, P, O
-
-                sp_po_batch[batch_index,] = keys[example_index]
+                prev_end = 0
+                for query_index, (query, end) in enumerate(self.query_end_index.items()):
```
rgemulla, line 605: Likewise, any order assumed?

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
+                    if example_index < end:
+                        example_index -= prev_end
+                        query_type[batch_index] = query_index
+                        keys = self.keys[query]
+                        offsets = self.offsets[query]
+                        values = self.values[query]
+                        if query_index == 0:
```
rgemulla, line 612: Woudln't it be more natural if query_index refers to the position of the element being predicted? (`0=_po, 1=s_o, 2=sp_`)?

kge/config-default.yaml
```diff
@@ -220,6 +220,12 @@ KvsAll:
...
+  queries:
+    sp_to_o: True
+    so_to_p: False
+    po_to_s: True
```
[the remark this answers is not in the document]
rgemulla: Perhaps we should change the naming to make everything more readable. E.g.

query_type: e.g., sp_, (use this in the config)
target: e.g., o (the _ part of a query_type)
query: an actual query, e.g. (AlbertEinstein, born_In)
labels: the result(s) of query 
example: a query + its labels

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
-            self.train_sp_keys,
-            self.train_sp_values,
-            self.train_sp_offsets,
-        ) = kge.indexing.prepare_index(train_sp)
-        (
-            self.train_po_keys,
-            self.train_po_values,
-            self.train_po_offsets,
-        ) = kge.indexing.prepare_index(train_po)
+        self.queries = self.config.get("KvsAll.queries")
```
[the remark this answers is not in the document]
rgemulla: Should be query_type with the suggested naming

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
-
         # create the collate function
         def collate(batch):
-            """For a batch of size n, returns a triple of:
+            """For a batch of size n, returns a dictionary of:
 
-            - pairs (nx2 tensor, row = sp or po indexes),
-            - label coordinates (position of ones in a batch_size x num_entities tensor)
-            - is_sp (vector of size n, 1 if corresponding example_index is sp, 0 if po)
+            - pairs (nx2 tensor, row = sp, po or so indexes),
```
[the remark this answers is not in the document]
rgemulla: pairs -> queries

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
+                        num_ones += self.offsets[query][example_index + 1]
+                        num_ones -= self.offsets[query][example_index]
+                        break
+                    prev_end = end
 
             # now create the results
-            sp_po_batch = torch.zeros([len(batch), 2], dtype=torch.long)
-            is_sp = torch.zeros([len(batch)], dtype=torch.long)
+            key_batch = torch.zeros([len(batch), 2], dtype=torch.long)
+            query_type = torch.zeros([len(batch)], dtype=torch.long)
```
rgemulla, line 599: query_types

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
+                prev_end = 0
+                for query_index, (query, end) in enumerate(self.query_end_index.items()):
+                    if example_index < end:
+                        example_index -= prev_end
+                        query_type[batch_index] = query_index
+                        keys = self.keys[query]
+                        offsets = self.offsets[query]
+                        values = self.values[query]
+                        if query_index == 0:
+                            query_col_1, query_col_2, value_col = S, P, O
```
rgemulla, line 613: target_col

kge/job/train.py
```diff
@@ -640,50 +650,61 @@ def collate(batch):
...
+            key_indexes[query] = (query_type == query_index).nonzero().to(self.device).view(-1)
+
         labels = kge.job.util.coord_to_sparse_tensor(
-            batch_size, self.dataset.num_entities(), label_coords, self.device
+            batch_size, max(self.dataset.num_entities(), self.dataset.num_relations()), label_coords, self.device
         ).to_dense()
+
+        query_labels = {}
+        for query, indexes in key_indexes.items():
+            if query == "so_to_p":
```
rgemulla, line 667: This should become query_type == "s_o"

kge/job/train.py
```diff
@@ -530,108 +530,118 @@ def __init__(self, config, dataset, parent_job=None):
...
-                if example_index < num_sp:
-                    num_ones += self.train_sp_offsets[example_index + 1]
-                    num_ones -= self.train_sp_offsets[example_index]
-                else:
-                    example_index -= num_sp
-                    num_ones += self.train_po_offsets[example_index + 1]
-                    num_ones -= self.train_po_offsets[example_index]
+                prev_end = 0
+                for query, end in self.query_end_index.items():
+                    if example_index < end:
```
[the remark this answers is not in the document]
vonVogelstein: relies on dict to be in some order (which is always the case in python)

kge/job/train.py
vonVogelstein: any

vonVogelstein: any

kge/job/train.py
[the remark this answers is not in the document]
rgemulla: Is it? While I do think that this is the case in any likely implementation as long as the keyset is not changed, making such an assumption makes the code hard to read. The value of "end" depends on the order of the keys in the iteration.

More readable may be to use a map from query to (begin,end) or an ordered array of (end_index, query_type) pairs. An ordereddict would also be fine, but that still makes the code somewhat unaccessible.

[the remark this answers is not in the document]
rgemulla: I had another look. The code sets `self.query_end_index[query] = prev_end + len(data[query])`. It implicitly assumes that `query_end_index` retains insertion order / is ordered the same way as `queries`. That's a hard-to-grasp assumption and it may not be valid.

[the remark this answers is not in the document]
vonVogelstein: As far as I can see insertion into `self.query_end_index` is based on a loop over `queries` and `self.query_end_index` is a dict, therefore retaining insertion order according to the order in queries. Anyhow, I will adjust the code to be easier to understand

vonVogelstein:
As discussed used new naming conventions and worked with list instead of dict where required

rgemulla:
Thanks! I pushed some further revisions into your branch. Now let's reverse our roles: Please check if everything is OK and you agree with these changes.

kge/job/train.py
```diff
@@ -530,37 +530,49 @@ def __init__(self, config, dataset, parent_job=None):
...
 
+        # determine enabled query types
+        self.query_types = [
+            key
+            for key, enabled in self.config.get("KvsAll.query_types").items()
+            if enabled
+        ]
+
+        #' for each query type: list of queries
+        self.queries = {}
```
vonVogelstein, line 543: definition of query should be given for easier understanding

kge/job/train.py
```diff
@@ -530,37 +530,49 @@ def __init__(self, config, dataset, parent_job=None):
...
+
+        #' for each query type: list of all labels (concatenated across queries)
+        self.labels = {}
```
vonVogelstein, line 546: definition of label should be given for easier understanding

kge/job/train.py
```diff
@@ -530,37 +530,49 @@ def __init__(self, config, dataset, parent_job=None):
...
+
+        #' for each query type: list of starting offset of labels in self.labels. The
+        #' labels for the i-th query of query_type are in labels[query_type] in range
+        #' label_offsets[query_type][i]:label_offsets[query_type][i+1]
+        self.label_offsets = {}
+
+        #' for each query type (ordered as in self.query_types), index of last example
+        #' of that type in the list of all examples
+        self.query_end_index = []
```
vonVogelstein, line 555: Is it not the index of the last query of that type in the list of all queries (according to new definitions)?
Also, it is not the index of the last query but `1+ index of last query`, therefore the end in terms of pythonic slicing

kge/job/train.py
```diff
@@ -619,85 +632,105 @@ def collate(batch):
...
+        query_type_indexes_batch = batch["query_type_indexes"]
+
+        examples_for_query_type = {}
+        for query_type_index, query_type in enumerate(self.query_types):
+            examples_for_query_type[query_type] = (
+                (query_type_indexes_batch == query_type_index)
+                .nonzero()
+                .to(self.device)
+                .view(-1)
+            )
```
vonVogelstein, line 683: should be `queries_for_query_type`

rgemulla:
Thanks! Any more comments on this?

vonVogelstein:
No, everything looks fine to me

rgemulla:
Thanks for your help!

PR #88, opened 2020-03-26
Allow short job_ids for the --job_id option in dump trace

Nzteb:
The current dump trace command expects a full job_id for the option --job_id to work. This update allows that specifying the short form of the job_id also works.


rgemulla:
Thanks!

Issue #90, opened 2020-04-09
Error when trying ax search


apoorvumang:
Hi

Training is working fine (train config in README). However when I try to do hyperparameter search using the config provided in README, I get following error:

```
Traceback (most recent call last):
  File "/scratche/home/apoorv/kge/env/bin/kge", line 11, in <module>
    load_entry_point('libkge', 'console_scripts', 'kge')()
  File "/scratche/home/apoorv/kge/kge/cli.py", line 282, in main
    raise e from None
  File "/scratche/home/apoorv/kge/kge/cli.py", line 278, in main
    job.run()
  File "/scratche/home/apoorv/kge/kge/job/auto_search.py", line 111, in run
    self.init_search()
  File "/scratche/home/apoorv/kge/kge/job/ax_search.py", line 43, in init_search
    model_kwargs={'seed': 0}
TypeError: __new__() got an unexpected keyword argument 'num_arms'
```

I created a folder 'wn' and added the config.yaml file. Then I ran` kge resume wn`
I have attached kge.log that was created.
Any help would be appreciated!
Thanks
Apoorv
[kge.log](https://github.com/uma-pi1/kge/files/4456960/kge.log)

rgemulla:
I updated the example in the README slightly so that it uses Complex (there was an unrelated issue with this particular example).

I just tried and the updated example works fine. If you still see this problem, what are your Pytorch and Ax versions?

apoorvumang:
I'm still seeing the same problem :(

```
>>> torch.__version__
'1.4.0'
>>> ax.__version__
'0.1.10'
```

apoorvumang:
I commented the line  `num_sobol_trials: 10` and the code works. Is at an issue with ax on my side?

AdrianKs:
Ax changed their parameter names with the latest version.
Ycou can compare the names here:
https://ax.dev/versions/0.1.8/api/_modules/ax/modelbridge/dispatch_utils.html
https://ax.dev/versions/latest/api/_modules/ax/modelbridge/dispatch_utils.html

I created a PR to fix that.

apoorvumang:
Thanks!

PR #91, opened 2020-04-10
update ax parameter name for latest ax version

AdrianKs:
ax changed their parameter names with the latest update. This should fix this issue: https://github.com/uma-pi1/kge/issues/90


AdrianKs:
This change is actually for latest and not yet for 0.1.9

PR #94, opened 2020-04-15
create packaged models from checkpoint

AdrianKs:
Allows to create a packaged model, which only contains the model, entity/relation ids and the config.
It is not possible to resume training on a packaged model but it can be evaluated.

example command:
kge package checkpoint_best.pt

normal checkpoints also store entity/relation ids now


kge/job/train.py
```diff
@@ -242,6 +242,8 @@ def save(self, filename) -> None:
                 "optimizer_state_dict": self.optimizer.state_dict(),
                 "lr_scheduler_state_dict": self.kge_lr_scheduler.state_dict(),
                 "job_id": self.job_id,
+                "entity_ids": self.dataset.entity_ids(),
+                "relation_ids": self.dataset.relation_ids(),
```
rgemulla, line 246: Should not be part of normal checkpoint (may increase size significantly).

kge/job/train.py
```diff
@@ -252,18 +254,21 @@ def load(self, filename: str) -> str:
         Returns job id of the job that created the checkpoint."""
         self.config.log("Loading checkpoint from {}...".format(filename))
         checkpoint = torch.load(filename, map_location="cpu")
+        if self.config.get("job.type") == "train" and checkpoint["type"] == "package":
+            raise ValueError("Can not continue training on a packaged model")
```
rgemulla, line 258: != checkpoint

kge/job/train.py
```diff
@@ -252,18 +254,21 @@ def load(self, filename: str) -> str:
...
         else:
             # old format (deprecated, will eventually be removed)
             self.model.load_state_dict(checkpoint["model_state_dict"])
-        self.optimizer.load_state_dict(checkpoint["optimizer_state_dict"])
-        if "lr_scheduler_state_dict" in checkpoint:
-            # new format
-            self.kge_lr_scheduler.load_state_dict(checkpoint["lr_scheduler_state_dict"])
-        self.epoch = checkpoint["epoch"]
-        self.valid_trace = checkpoint["valid_trace"]
+        if checkpoint["type"] != "package":
```
rgemulla, line 265: This should always be the case here.

rgemulla:
How are pacakged models used? Maybe we should rename ``KgeModel.load_from_checkpoint`` to ``KgeModel.load`` and support checkpoints and packages (including creation of required "virtual" datasets) there.

rgemulla:
TrainingJob.load should offload some functionality to ``KgeModel.load_from``. And I guess ``KgeModel.load_from`` should accept both filenames and loaded files (to facilitate this).

kge/job/train.py
AdrianKs: currently default type name is "train". Since this is in all old checkpoints, I would just leave it like this.

kge/job/train.py
AdrianKs: the trainjob.load method is also used if we call kge eval/test.
therefore we need this line to be able to load packaged models for evaluation and don't get complications with the optimizer, which we don't need

AdrianKs:
> TrainingJob.load should offload some functionality to `KgeModel.load_from`. And I guess `KgeModel.load_from` should accept both filenames and loaded files (to facilitate this).

This could be done, but I am not sure if we should do it.
KgeModel.load_from creates a new model based on the config file provided in the checkpoint.
At the time we call TrainingJob.load we already have a model created which we just update with the parameters in the checkpoint. 

If we want to move this to the load_from function we would also need to provide the new config file and there we unnecessary create a new model.
We would only offload the functionality of the model loading to load_from

This creates an overhead with no big advantages.

kge/job/train.py
rgemulla: We shouldn't use TJ.load for eval/test jobs: there is no need to load things like optimizer states. And making training jobs aware of other jobs is not a good design.

kge/job/train.py
rgemulla: Ok, then != train

rgemulla:
I see. We still need load_from to work with packages (that's a key point of having packages). And we should try to minimize code duplication. Ideas welcome!

AdrianKs:
This does not help much with the code duplication but with the designflaw having the evaluation job use the trainjob resume.
lets move most of the content from trainjob.resume to the new method job.resume.
In job.resume we check which checkpoint we need to load and load it.

trainjob.resume then looks something like this:
`
def resume(self, checkpoint_file)  \
   checkpoint = super().resume(checkpoint_file) \
    self.load(check_point)
`
evaljob.resume then just looks the same
but the load functions can be different

this does not help much with the load and load_from distinction but in general would be a lot cleaner, even if there is maybe even more duplication.
but the train job types are more independent.

AdrianKs:
actually also the self.load(checkpoint) call could be done in job.resume
then we only have a job specific load function.

rgemulla:
Sounds good. But note that resuming from a checkpoint generally may need to use the config from the checkpoint, not the one from the folder (which may not even exist). When cleaning up the API, this should be considered. We'd need sth. like TrainingJob.load_from(checkpoint) and KgeModel.load_from(checkpoint) as static methods. One advantage is that the training job then does not need to create a model first, it simply calls KgeModel.load_from. Also, the methods should accept both a filename or a loaded checkpoint in their "checkpoint" argument. This way, we'd only need to load the checkpoint once.

AdrianKs:
in that case the function resume() could maybe completly be replaced by the method static method load_from, since we create a new model and then call resume and afterwards run() anyways currently this could be repaced by job = load_from() checkpoint and afterwards job.run().

rgemulla:
Sounds like a good idea to me.

AdrianKs:
I built a first version to replace the method resume with load_from. This actually resulted in more changes than expected. Please have a look, if this is worth the changes.

AdrianKs:
With these changes we could also make it possible to evaluate checkpoints without the config, I think

kge/job/eval.py
```diff
@@ -83,19 +87,30 @@ def run(self) -> dict:
...
+        parent_job=None,
+        device="cpu",
+    ) -> Job:
+        checkpoint_file = None
+        if checkpoint is None or type(checkpoint) is str:
+            checkpoint_file = checkpoint
+            checkpoint, config = super().load_from(
+                checkpoint, config=config, device=device
+            )
+        model = KgeModel.load_from(checkpoint, config=config, dataset=dataset)
```
rgemulla, line 105: Part of super?

kge/job/grid_search.py
```diff
@@ -68,8 +68,12 @@ def run(self):
 
         # and run it
         if self.config.get("grid_search.run"):
-            job = Job.create(self.config, self.dataset, parent_job=self)
-            job.resume()
+            job = JobFactory.load_from(
+                checkpoint_file=None,
+                config=self.config,
+                dataset=self.dataset,
```
rgemulla, line 74: Where does the dataset come from?

kge/job/job.py
```diff
@@ -39,7 +40,7 @@ class Job:
         _save_job_config,
     ]
 
-    def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
+    def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None, init=True):
```
rgemulla, line 43: Why is this needed?

kge/job/job.py
```diff
@@ -55,36 +56,29 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
-        use :func:`run`.
-
-        Should set `resumed_from_job` to the job ID of the previous job.
-
-        """
-        raise NotImplementedError
+    @staticmethod
+    def load_from(checkpoint_file: str = None, config: Config = None, device="cpu", parent_job=None) -> (Dict, Optional[Config]):
+        if checkpoint_file is None and config is None:
+            raise ValueError("Either a config file or a checkpoint file need to be provided to load a checkpoint.")
```
rgemulla, line 62: Confusing message. How can a Job be loaded without a checkpoint?

kge/job/job_factory.py
```diff
@@ -0,0 +1,41 @@
...
+from typing import Optional
+
+
+class JobFactory:
+    @staticmethod
+    def pick_job(job_type=None, config=None):
+        from kge.job import TrainingJob, EvaluationJob, SearchJob
+
+        if job_type is None and config is None:
+            job_type = "train"
```
rgemulla, line 12: Do not default to anything here.

kge/job/job_factory.py
```diff
@@ -0,0 +1,41 @@
...
+    @staticmethod
+    def create(config: Config, dataset: Dataset, parent_job=None, init=True):
+        job_class = JobFactory.pick_job(config=config)
+        return job_class.create(config, dataset, parent_job=parent_job, init=init)
+
+    @staticmethod
+    def load_from(
+        checkpoint_file: str,
+        job_type=None,
+        config: Optional[Config] = None,
```
rgemulla, line 34: What's the semantic of this config? Does it override the one from the checkpoint? If so, that would be a good design, imho. We can then, for example, override the job type with a new config that only holds "job.type=x". The job_type argument is then also not needed (one may also use EvalJob.load_from directly).

kge/job/job_factory.py
```diff
@@ -0,0 +1,41 @@
...
+        dataset: Optional[Dataset] = None,
```
rgemulla, line 35: What happens if dataset is set to None?

kge/job/search.py
```diff
@@ -11,7 +11,7 @@ class SearchJob(Job):
     Provides functionality for scheduling training jobs across workers.
     """
 
-    def __init__(self, config, dataset, parent_job=None):
+    def __init__(self, config, dataset, parent_job=None, init=True):
```
rgemulla, line 14: Why the init=True here (and elsewhere)?

kge/job/train.py
```diff
@@ -37,33 +38,27 @@ class TrainingJob(Job):
...
-        )
+
         self.config.check("train.trace_level", ["batch", "epoch"])
         self.trace_batch: bool = self.config.get("train.trace_level") == "batch"
         self.epoch: int = 0
         self.valid_trace: List[Dict[str, Any]] = []
         self.is_prepared = False
-        self.model.train()
+        if init:
+            self._initialize_training(config, dataset)
```
rgemulla, line 61: Why is this not always done? An easier to read variant may be to pass "checkpoint" instead of init. The code can then decide what to do.

kge/job/train.py
```diff
@@ -98,17 +93,36 @@ def __init__(
...
+        self.kge_lr_scheduler = KgeLRScheduler(config, self.optimizer)
+        valid_conf = config.clone()
+        valid_conf.set("job.type", "eval")
+        if self.config.get("valid.split") != "":
+            valid_conf.set("eval.split", self.config.get("valid.split"))
+        valid_conf.set("eval.trace_level", self.config.get("valid.trace_level"))
+        self.valid_job = EvaluationJob.create(
+            valid_conf, dataset, parent_job=self, model=self.model
+        )
+        self.model.train()
```
rgemulla, line 113: This last line should probably always be done.

kge/model/kge_model.py
```diff
@@ -409,7 +415,14 @@ def load_from_checkpoint(
             if not config.log_folder:
                 config.log_folder = "."
         if dataset is None:
-            dataset = Dataset.load(config, preload_data=False)
+            if checkpoint["type"] == "package":
+                dataset = Dataset(config)
+                dataset._meta["entity_ids"] = checkpoint["entity_ids"]
```
rgemulla, line 420: I suggest to make this more generic. dataset.save_meta(cp, keys) and dataset.load_meta(cp). Store under "dataset.meta.xxx".

rgemulla:
I think it's worth it. Perhaps some of the more generic checkpoint-handling code can be put into a seaparate package (kge.util.io?).

In general, I think we should have KgeModel.load_from_checkpoint(cp), Job.load_from_checkpoint(cp), Dataset.load_from_checkpoint(cp), and Config.load_from_checkpoint(cp). If we also had the reverse methods (save_to_checkpoint) everywhere, we could remove code clutter and code duplication.

kge/cli.py
```diff
@@ -272,9 +279,10 @@ def main():
             dataset = Dataset.load(config)
 
             # let's go
-            job = Job.create(config, dataset)
             if args.command == "resume":
-                job.resume(checkpoint_file)
+                job = JobFactory.load_from(checkpoint_file, config=config)
```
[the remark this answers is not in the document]
AdrianKs: moved back to Job. 
config is optional, default is config from checkpoint

kge/job/eval.py
```diff
@@ -83,19 +87,30 @@ def run(self) -> dict:
...
-        training_job.resume(checkpoint_file)
-        self.model = training_job.model
-        self.epoch = training_job.epoch
-        self.resumed_from_job_id = training_job.resumed_from_job_id
-        self.trace(
-            event="job_resumed", epoch=self.epoch, checkpoint_file=checkpoint_file
+    @classmethod
+    def load_from(
+        cls,
+        checkpoint: Union[str, Dict] = None,
```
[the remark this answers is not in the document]
AdrianKs: this was to handle functionality from old resume function.
I created a new resume function now to handle the selection of the correct checkpoint, if only a config is provided

kge/job/eval.py
```diff
@@ -83,19 +87,30 @@ def run(self) -> dict:
...
+        config: Config = None,
+        dataset: Dataset = None,
+        parent_job=None,
+        device="cpu",
+    ) -> Job:
+        checkpoint_file = None
+        if checkpoint is None or type(checkpoint) is str:
+            checkpoint_file = checkpoint
+            checkpoint, config = super().load_from(
```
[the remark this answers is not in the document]
AdrianKs: fixed

kge/job/eval.py
AdrianKs: fixed

kge/job/grid_search.py
AdrianKs: it is now created either, when the job is created or resumed

kge/job/search.py
AdrianKs: this made it possible to not create a model on init of training job. Added everywhere else to keep signature persistent.
Now replaced with model parameter to directly set the model. Only added to training job signature

kge/job/train.py
AdrianKs: fixed

kge/model/kge_model.py
AdrianKs: fixed

kge/model/embedder/lookup_embedder.py
```diff
@@ -58,7 +57,7 @@ def __init__(
                 dropout = 0
         self.dropout = torch.nn.Dropout(dropout)
 
-    def prepare_job(self, job: Job, **kwargs):
```
AdrianKs, line 61: I had to remove the Job dependency, due to circular imports (I had to add KgeModel to Job). Do you have a better idea than just to remove it?
Dependency Injection seems a little overkill for the little type annotation.

AdrianKs:
I tried to address most of your points.

kge/config.py
```diff
@@ -423,6 +425,26 @@ def init_folder(self):
             return True
         return False
 
+    @staticmethod
+    def load_from(checkpoint_config: Config, new_config=None, folder=None) -> Config:
```
rgemulla, line 429: Rename new_config to override_config to make it mode clear; add type

kge/cli.py
```diff
@@ -272,9 +279,10 @@ def main():
             dataset = Dataset.load(config)
 
             # let's go
-            job = Job.create(config, dataset)
             if args.command == "resume":
-                job.resume(checkpoint_file)
+                job = Job.resume(checkpoint_file, config=config)
```
rgemulla, line 283: dataset not used here (generally, dataset argument should be optional)

kge/cli.py
```diff
@@ -272,9 +279,10 @@ def main():
...
+            else:
+                job = Job.create(config, dataset)
```
rgemulla, line 285: If it's optional, we do not use to compute it here at all

kge/dataset.py
```diff
@@ -91,6 +93,20 @@ def load(config: Config, preload_data=True):
                 dataset.split(split)
         return dataset
 
+    @staticmethod
+    def load_meta(config: Config, meta_checkpoint: Dict, preload_data=True) -> Dataset:
```
rgemulla, line 97: For loading, just use load_from.

kge/dataset.py
```diff
@@ -91,6 +93,20 @@ def load(config: Config, preload_data=True):
...
+        """Loads a dataset and overrides meta data."""
+        dataset = Dataset.load(config, preload_data)
+        dataset._meta.update(meta_checkpoint)
+        return dataset
+
+    def save_meta(self, keys: List[str]):
```
rgemulla, line 103: Use save_to_checkpoint as well, but accept a meta argumetn. If not given, nothing is saved.

kge/job/auto_search.py
```diff
@@ -49,24 +49,16 @@ def save(self, filename):
...
-        if checkpoint_file is not None:
-            self.resumed_from_job_id = self.load(checkpoint_file)
-            self.trace(
-                event="job_resumed", checkpoint_file=checkpoint_file
-            )
-            self.config.log(
-                "Resumed from {} of job {}".format(
-                    checkpoint_file, self.resumed_from_job_id
-                )
+    def load(self, checkpoint):
```
rgemulla: We should try to use consistent names (load_from, save_to)

kge/job/eval.py
```diff
@@ -83,20 +87,27 @@ def run(self) -> dict:
...
+    def load_from(
+        cls,
+        checkpoint: Union[str, Dict],
+        config: Config = None,
+        dataset: Dataset = None,
+        parent_job=None,
+    ) -> Job:
+        if config is None:
+            config = Config()
+            config.set("job.type", "eval")
```
rgemulla, line 107: Not necessary

kge/job/job.py
```diff
@@ -55,36 +59,84 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
-        raise NotImplementedError
-
-    def run(self):
-        raise NotImplementedError
-
-    def create(config: Config, dataset: Dataset, parent_job: "Job" = None) -> "Job":
-        """Creates a job for a given configuration."""
-
+    @staticmethod
+    def create(config: Config, dataset: Dataset, parent_job=None, model=None):
```
rgemulla, line 63: dataset should be optional

kge/job/search.py
```diff
@@ -130,8 +131,12 @@ def _run_train_job(sicnk, device=None):
                 train_job_config.get("job.device"),
             )
         )
-        job = Job.create(train_job_config, search_job.dataset, parent_job=search_job)
-        job.resume()
+        job = Job.resume(
```
rgemulla, line 134: Confusing that this is called resume (and not create). It does not actually resume anything, but loads a job. So why not load_from?

kge/job/train.py
```diff
@@ -246,45 +252,55 @@ def save(self, filename) -> None:
             filename,
         )
 
-    def load(self, filename: str) -> str:
+    def load(self, checkpoint: Dict, model: Optional[KgeModel] = None) -> str:
```
rgemulla, line 255: Hmm. Perhaps we should use "load" for instance methods and "create" / "create_from" for class methods? Would this make it clearer?

kge/model/embedder/lookup_embedder.py
rgemulla: Perhaps you can import KgeModel in Job only where needed?

kge/model/kge_model.py
```diff
@@ -393,23 +398,24 @@ def load_from_checkpoint(
         appended to) in the checkpoint's folder.
 
         """
+        if type(checkpoint) is str:
+            checkpoint = load_checkpoint(checkpoint, device=device)
 
-        checkpoint = torch.load(filename, map_location=device)
+        config = Config.load_from(checkpoint["config"], config)
```
rgemulla, line 404: This should just be: config.load_from(checkpoint, override=config) or so. Which keys to use is up to config / dataset / train / etc.

kge/model/kge_model.py
```diff
@@ -393,23 +398,24 @@ def load_from_checkpoint(
...
-            config.log_folder = os.path.dirname(filename)
+            config.log_folder = checkpoint["folder"]
             if not config.log_folder:
                 config.log_folder = "."
         if dataset is None:
-            dataset = Dataset.load(config, preload_data=False)
+            if checkpoint["type"] == "package":
+                dataset.load_meta(config, checkpoint["dataset_meta"], preload_data=False)
+            else:
+                dataset = Dataset.load(config, preload_data=False)
```
rgemulla, line 418: Dataset should always load what's in the checkpoint (i.e., if there is meta data load it). We probably also want to put num_entities and num_relations into the checkpoint.

rgemulla: Use dataset.meta

kge/util/package.py
```diff
@@ -0,0 +1,33 @@
...
+
+def package_model(checkpoint_path):
+    """
+    Converts a checkpoint to a packaged model.
+    A packaged model only contains the model, entity/relation ids and the config.
+    """
+    checkpoint = torch.load(checkpoint_path, map_location="cpu")
+    original_config = checkpoint["config"]
+    config = Config()  # round trip to handle deprecated configs
+    config.load_options(original_config.options)
```
rgemulla, line 23: Config.load_from

kge/util/package.py
```diff
@@ -0,0 +1,33 @@
...
+    dataset = Dataset.load(config, preload_data=False)
+    packaged_model = {
```
rgemulla, line 25: Create an empty dict. Then call config.save_to(cp), dataset.save_to(cp, meta=...) etc.

kge/util/package.py
```diff
@@ -0,0 +1,33 @@
...
+        "type": "package",
+        "model": checkpoint["model"],
+        "config": checkpoint["config"],
+        "dataset_meta": dataset.save_meta(["entity_ids", "relation_ids", "entity_strings", "relation_strings"])
+    }
+    output_folder = os.path.dirname(checkpoint_path)
+    filename = "model_{}.pt".format(checkpoint["epoch"])
```
rgemulla, line 32: perhaps replace "checkpoint" in filename by "model"? And if not present, add "_package"? Should also be possible to specifiy as command line option.

AdrianKs: done

kge/util/package.py
AdrianKs: done

kge/util/package.py
AdrianKs: now called create_from

kge/model/kge_model.py
AdrianKs: done

kge/model/kge_model.py
AdrianKs: now called create_from

kge/job/train.py
AdrianKs: done

kge/job/search.py
AdrianKs: now called find_and_create. If you have a better idea for the name, let me know

kge/job/job.py
AdrianKs: done

kge/job/eval.py
AdrianKs: removed

kge/dataset.py
AdrianKs: done, now called save_to

kge/dataset.py
AdrianKs: now called create_from

kge/cli.py
AdrianKs: fixed

kge/cli.py
AdrianKs: fixed

kge/config.py
AdrianKs: done

kge/model/kge_model.py
AdrianKs: I did not yet add num_entities and num_relations.
Do we want to have that only in the packaged model, but also in the train checkpoints?

rgemulla: It should always be added in save_to for the Dataset, I guess. But we do not need to call this for training checkpoints. Generally, the datasets create_from should take whatever it finds in the checkpoint.

BTW: Please resolve all conversation which are done.

kge/job/auto_search.py
AdrianKs: instance methods now called load and static/class methods create_from

kge/model/embedder/lookup_embedder.py
AdrianKs: done

kge/job/train.py
AdrianKs: removed

kge/job/job_factory.py
AdrianKs: job factory removed

kge/job/job.py
AdrianKs: removed

kge/job/job_factory.py
AdrianKs: can be used without a config now.
If we call EvalJob.load_from() directly without any config, it will currently not know what type of job to create

rgemulla:
I think valid is a sensible default choice.

AdrianKs:
It is now possible to evaluate without a config by calling EntityRankingJob.create_from(checkpoint).

If we later want to change the console api to enable evaluation without a config, we would have to find a way to still support additional commandline options without overwriting the checkpoint config with all the default values

rgemulla:
Is this PR ready for another review?

AdrianKs:
yes it is

kge/job/eval.py
```diff
@@ -83,20 +87,33 @@ def run(self) -> dict:
...
-        from kge.job import TrainingJob
-
-        training_job = TrainingJob.create(self.config, self.dataset)
-        training_job.resume(checkpoint_file)
-        self.model = training_job.model
-        self.epoch = training_job.epoch
-        self.resumed_from_job_id = training_job.resumed_from_job_id
+    def load(self, checkpoint, model):
+        if checkpoint["type"] == "search":
+            raise ValueError("Can not evaluate on a search job checkpoint.")
```
rgemulla, line 92: I'd do this the other way around: only allow train and package checkpoints.

kge/job/eval.py
```diff
@@ -83,20 +87,33 @@ def run(self) -> dict:
...
+        self.epoch = checkpoint["epoch"]
         self.trace(
-            event="job_resumed", epoch=self.epoch, checkpoint_file=checkpoint_file
+            event="job_resumed", epoch=self.epoch, checkpoint_file=checkpoint["file"]
         )
 
+    @classmethod
+    def create_from(
+        cls,
+        checkpoint: Union[str, Dict],
```
rgemulla, line 102: Inconsistent: either allow Union[str,Dict] everywhere or nowhare

kge/job/eval.py
```diff
@@ -83,20 +87,33 @@ def run(self) -> dict:
...
+        config: Config = None,
```
rgemulla, line 103: override_config?

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+        elif job_type == "eval":
+            return EvaluationJob.create(
+                config, dataset, parent_job=parent_job, model=model
+            )
+        else:
+            raise ValueError("unknown job type")
 
-        Should set `resumed_from_job` to the job ID of the previous job.
+    @staticmethod
+    def find_and_create_from(
```
rgemulla, line 85: Why not just create_from?

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+            dataset: dataset object
+            parent_job: parent job (e.g. search job)
+
+        Returns: Job based on checkpoint or new Job if no checkpoint found
 
         """
-        raise NotImplementedError
+        if checkpoint is None and config is None:
+            raise ValueError(
+                "Please provide either the config file located in the folder structure "
```
rgemulla, line 107: Confusing: config may not be in a "folder structure". Just say: config or checkpoint requried.

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+                "containing the checkpoint or the checkpoint itself."
+            )
+        elif checkpoint is None:
```
rgemulla, line 110: elif -> if (since after raise)

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+            if last_checkpoint is not None:
+                checkpoint = config.checkpoint_file(last_checkpoint)
+
+        if checkpoint is not None:
+            job = Job.create_from(checkpoint, config, dataset, parent_job=parent_job)
+            if type(checkpoint) == str:
+                job.config.log("Loading checkpoint from {}...".format(checkpoint))
+        else:
+            job = Job.create(config, dataset, parent_job=parent_job)
+            job.config.log("No checkpoint found, starting from scratch...")
```
rgemulla, line 121: found or specified

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+            )
+        elif checkpoint is None:
+            last_checkpoint = config.last_checkpoint()
+            if last_checkpoint is not None:
+                checkpoint = config.checkpoint_file(last_checkpoint)
+
+        if checkpoint is not None:
+            job = Job.create_from(checkpoint, config, dataset, parent_job=parent_job)
+            if type(checkpoint) == str:
+                job.config.log("Loading checkpoint from {}...".format(checkpoint))
```
rgemulla, line 118: else: loaded checkpoint

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
 
-    def run(self):
-        raise NotImplementedError
+    @classmethod
+    def create_from(
+        cls,
+        checkpoint: Union[str, Dict],
+        config: Config = None,
+        dataset: Dataset = None,
+        parent_job=None,
```
rgemulla, line 130: Is it really required to have create_from and find_and_create_from and create?

kge/job/job.py
```diff
@@ -55,36 +58,115 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+    ) -> Job:
+        """
+        Creates a Job based on a checkpoint
+        Args:
+            checkpoint: path to checkpoint file or loaded checkpoint
+            config: config object
```
rgemulla, line 136: doc should say optional and semections (override_config?)

kge/job/train.py
```diff
@@ -232,59 +238,59 @@ def run(self) -> None:
...
         """Load job state from specified file.
 
         Returns job id of the job that created the checkpoint."""
-        self.config.log("Loading checkpoint from {}...".format(filename))
-        checkpoint = torch.load(filename, map_location="cpu")
-        if "model" in checkpoint:
-            # new format
-            self.model.load(checkpoint["model"])
+        if checkpoint["type"] != "train":
+            raise ValueError("Can not continue training on a packaged model")
```
rgemulla, line 266: Traning can only be continued on trainined checkpoints

kge/model/kge_model.py
```diff
@@ -393,23 +398,20 @@ def load_from_checkpoint(
...
-        config.load_options(original_config.options)
-        config.set("job.device", device)
         if use_tmp_log_folder:
             import tempfile
 
             config.log_folder = tempfile.mkdtemp(prefix="kge-")
         else:
-            config.log_folder = os.path.dirname(filename)
+            config.log_folder = checkpoint["folder"]
             if not config.log_folder:
```
rgemulla, line 412: or if it does not exsit

kge/util/package.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def package_model(checkpoint_path: str, filename: str = None):
+    """
+    Converts a checkpoint to a packaged model.
+    A packaged model only contains the model, entity/relation ids and the config.
+    """
+    checkpoint = load_checkpoint(checkpoint_path, device="cpu")
+    if checkpoint["type"] == "search":
+        raise ValueError("Can not package a search job.")
```
rgemulla, line 25: only allow train

kge/util/package.py
```diff
@@ -0,0 +1,44 @@
...
+    config = Config.create_from(checkpoint)
+    dataset = Dataset.create_from(checkpoint, config, preload_data=False)
+    packaged_model = {
+        "type": "package",
+        "model": checkpoint["model"],
+    }
+    packaged_model = config.save_to(packaged_model)
+    packaged_model = dataset.save_to(
+        packaged_model,
+        ["entity_ids", "relation_ids", "entity_strings", "relation_strings"],
```
rgemulla, line 35: only ids needed

rgemulla: more may be specified as arguments to package (not for this PR, only when it turns out we need this)

kge/config.py
```diff
@@ -423,6 +437,29 @@ def init_folder(self):
...
+        Overwrite options in checkpoint config with a new config.
+        Args:
+            checkpoint: loaded checkpoint
+            overwrite_config: new config with options to overwrite
+
+        Returns: Config object
+
+        """
+        config = Config()  # round trip to handle deprecated configs
+        if "config" in checkpoint and checkpoint["config"] is not None:
```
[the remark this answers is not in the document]
AdrianKs: Search checkpoints do not have a config file

kge/config.py
```diff
@@ -423,6 +437,29 @@ def init_folder(self):
...
+            config.load_options(checkpoint["config"].options)
+        if "folder" in checkpoint and checkpoint["folder"] is not None:
+            config.folder = checkpoint["folder"]
```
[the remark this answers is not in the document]
AdrianKs: checkpoint["folder"] is the checkpoint folder. I set this variable after reading the checkpoint.

kge/job/auto_search.py
```diff
@@ -49,24 +49,16 @@ def save(self, filename):
...
+        self.resumed_from_job_id = checkpoint.get("job_id")
+        self.trace(
+            event="job_resumed", checkpoint_file=checkpoint["file"]
+        )
+        self.config.log(
+            "Resumed from {} of job {}".format(
+                checkpoint["file"], self.resumed_from_job_id
             )
-        else:
-            self.config.log("No checkpoint found, starting from scratch...")
```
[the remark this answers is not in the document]
AdrianKs: search jobs are also continued with find_and_create_from. There it is logged. I think that is enough.

kge/job/eval.py
```diff
@@ -83,20 +87,33 @@ def run(self) -> dict:
...
-        """Load model state from last or specified checkpoint."""
-        # load model
-        from kge.job import TrainingJob
-
-        training_job = TrainingJob.create(self.config, self.dataset)
-        training_job.resume(checkpoint_file)
-        self.model = training_job.model
-        self.epoch = training_job.epoch
-        self.resumed_from_job_id = training_job.resumed_from_job_id
+    def load(self, checkpoint, model):
```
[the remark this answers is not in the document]
AdrianKs: yes, it is documented in job.py

kge/job/job.py
AdrianKs: we could merge create_from and find_and_create_from, but then check, but then the checkpoint parameter would be Optional[Union[str, Dict]].

kge/dataset.py
```diff
@@ -91,6 +93,53 @@ def load(config: Config, preload_data=True):
...
+            checkpoint: loaded checkpoint
+            config: config created from checkpoint
+            dataset: dataset to update
+            preload_data: preload data
+
+        Returns: created/updated dataset
+
+        """
+        if dataset is None:
+            dataset = Dataset.load(config, preload_data)
```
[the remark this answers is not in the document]
AdrianKs: ah, you are right, this does not work anymore. We would need some parameter to check if the created dataset is a mocked dataset, so that it throws an error, when we create a job.
Should I just add a parameter "mocked" to dataset or do you have a better idea, how I could handle that in job.create_from without just reloading the dataset?

rgemulla:
> we could merge create_from and find_and_create_from, but then check, but then the checkpoint parameter would be Optional[Union[str, Dict]].
Sounds good to me.

rgemulla:
> ah, you are right, this does not work anymore. We would need some parameter to check if the created dataset is a mocked dataset, so that it throws an error, when we create a job.
> Should I just add a parameter "mocked" to dataset or do you have a better idea, how I could handle that in job.create_from without just reloading the dataset? 
I think we shoul djust create the "normal" dataset (including setting the folders), but preload the data structures from the package. This way, everything works as long as nothing that's not packaged is accessed. Would this work?

AdrianKs:
> ah, you are right, this does not work anymore. We would need some parameter to check if the created dataset is a mocked dataset, so that it throws an error, when we create a job. Should I just add a parameter "mocked" to dataset or do you have a better idea, how I could handle that in job.create_from without just reloading the dataset?
> I think we shoul djust create the "normal" dataset (including setting the folders), but preload the data structures from the package. This way, everything works as long as nothing that's not packaged is accessed. Would this work?

I think this could break for evaluation jobs. If we load a packaged checkpoint without the dataset and then evaluate, it will fail. It would be nice if we could somehow catch that beforehand when we create the evaluation job.

rgemulla:
Why will it fail? If the dataset is present in the dataset folder, it will load the relevant data on demand. If the folder is not present and data should be loaded, we can raise an error "Dataset not found". If the specific data file is not present (e.g., a split), we can also raise the appropriate error.

AdrianKs:
the loading of the dataset should now work as expected. In KgeModel.create_from we just always set preload_false. In that way we can create a model without the dataset. If the datasets are later on needed it will throw an IOError

with that everything should be adressed

kge/cli.py
```diff
@@ -163,6 +165,11 @@ def main():
         dump(args)
         exit()
 
+    # package command
+    if args.command == "package":
+        package_model(args.checkpoint, args.file)
```
rgemulla, line 170: Give entire args as argument (as with dump above).

kge/dataset.py
```diff
@@ -91,6 +93,53 @@ def load(config: Config, preload_data=True):
                 dataset.split(split)
         return dataset
 
+    @staticmethod
+    def create_from(
+        checkpoint: Dict,
+        config: Config,
+        dataset: Optional[Dataset] = None,
+        preload_data=True,
```
rgemulla, line 101: Perhaps better to use preload=False as default. This makes it easier to use packages. I am actually not sure if we should support preloading at all.

kge/dataset.py
```diff
@@ -91,6 +93,53 @@ def load(config: Config, preload_data=True):
...
+        dataset_checkpoint = {
+            "num_entities": self.num_entities(),
+            "num_relations": self.num_relations(),
+        }
+        checkpoint["dataset"] = dataset_checkpoint
+        if meta_keys is None:
+            return checkpoint
+        meta_checkpoint = {}
+        for key in meta_keys:
+            meta_checkpoint[key] = self._map_indexes(None, key)
```
rgemulla, line 139: _map_indexes seems to be incorrect here. Do you mean map_indexes? Also, why not just store meta_["key"]?

kge/dataset.py
```diff
@@ -117,6 +166,10 @@ def _load_triples(filename: str, delimiter="\t", use_pickle=False) -> Tensor:
     def load_triples(self, key: str) -> Tensor:
         "Load or return the triples with the specified key."
         if key not in self._triples:
+            if self.folder is None or not os.path.exists(self.folder):
```
rgemulla, line 169: add a method `is_available` or so and move the code there. This way, we can add more checks later on. E.g., is_available() may check everything, `is_available(key)` may check an individual file etc.

kge/job/eval.py
```diff
@@ -83,20 +87,36 @@ def run(self) -> dict:
...
```
rgemulla, line 90: Add types

kge/job/eval.py
```diff
@@ -83,20 +87,36 @@ def run(self) -> dict:
...
         )
 
+    @classmethod
+    def create_from(
+        cls,
+        checkpoint: Dict,
+        overwrite_config: Config = None,
+        dataset: Dataset = None,
+        parent_job=None,
+        eval_split: str = "valid",
```
rgemulla, line 106: Drop this argument. It's not always used. If the split should be set, if should be done in the overwrite_config.

kge/job/job.py
```diff
@@ -55,36 +58,92 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
-        use :func:`run`.
+    @staticmethod
+    def create(
+        config: Config, dataset: Optional[Dataset] = None, parent_job=None, model=None
+    ):
+        from kge.job import TrainingJob, EvaluationJob, SearchJob
 
-        Should set `resumed_from_job` to the job ID of the previous job.
+        if dataset is None:
+            dataset = Dataset.load(dataset)
```
rgemulla, line 68: This looks shaky. What are we loading if dataset is None?

kge/job/job.py
```diff
@@ -55,36 +58,92 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+        checkpoint: Optional[Union[str, Dict]] = None,
+        overwrite_config: Config = None,
+        dataset: Dataset = None,
+        parent_job=None,
+    ) -> Job:
         """
-        raise NotImplementedError
+        Creates a Job based on a checkpoint
+        Args:
+            checkpoint: path to checkpoint file or loaded checkpoint
```
rgemulla, line 95: Why can't we just use a loaded checkpoint? I think we should have a utility method that takes a filename or a loaded checkpoint and returns the loaded checkpoint.

kge/job/job.py
```diff
@@ -55,36 +58,92 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
...
+
+        if overwrite_config is not None and overwrite_config.exists("job.device"):
+            device = overwrite_config.get("job.device")
         else:
-            raise ValueError("unknown job type")
-
+            device = "cpu"
+        if type(checkpoint) == str:
+            checkpoint = load_checkpoint(checkpoint, device)
+        overwrite_config = Config.create_from(checkpoint, overwrite_config)
```
rgemulla, line 125: Call lhs config?

kge/model/kge_model.py
```diff
@@ -380,10 +381,14 @@ def create_default(
         return model
 
     @staticmethod
-    def load_from_checkpoint(
-        filename: str, dataset=None, use_tmp_log_folder=True, device="cpu"
+    def create_from(
+        checkpoint: Union[str, Dict],
```
rgemulla, line 385: Also here: why not just Dict

kge/model/kge_model.py
```diff
@@ -380,10 +381,14 @@ def create_default(
...
+        dataset: Optional[Dataset] = None,
+        use_tmp_log_folder=True,
+        device="cpu",
+        config: Config = None,
```
rgemulla, line 389: overwrite_config? the config could alse have the device

kge/dataset.py
AdrianKs: I switched to map_indexes. If we just store meta["key"], we can not make sure, that we actually loaded this data. Could be that we didn't even read the entitiy_ids file, when calling this method.

kge/job/eval.py
AdrianKs: I added this argument, so that the user can easily evaluate without a config. With that he can just call EntityRankingJob.create_from(checkpoint_file, eval_split="test") to evaluate on test.
If we remove it, he has to create a complete empty config file with only the eval split specified.

kge/job/job.py
AdrianKs: ah, supposed to be config

kge/job/job.py
AdrianKs: just to make sure, before I restructure: we want to load the checkpoint before we call create_from?
then we would have to extend the method load_checkpoint to the signature load_checkpoint(checkpoint_file: Optional[str] = None, config: Optional[Config] = None). 
This method then loads the checkpoint_file, if provided or looks for the correct checkpoint, if only the config is provided.

rgemulla:
What's the current state here? Can you rebase this off the current master?

kge/dataset.py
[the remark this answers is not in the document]
rgemulla: How was this resolved?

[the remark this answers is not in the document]
AdrianKs: if preload_data is set to False, it will only create the dataset object without loading anything

kge/job/job.py
rgemulla: We can't we just pass a loaded checkpoint (or None) instead of Union[str,Dict]? Why does load_checkpoint require the config and not just the filename?

BTW: I think we should use clearer arguments/variable names throughout: `checkpoint` (loaded), `checkpoint_file` (filename), `checkpoint_or_file` (union) or so.

kge/job/eval.py
AdrianKs: documented overwrite priority

kge/job/job.py
AdrianKs: checkpoint loading is now completely separated from job.create_from

AdrianKs:
separated loading of checkpoint from create_from and rebased.
Should be done now.

rgemulla:
I made some revisions. Let me know your thoughts. If you are fine, we can merge this PR.

AdrianKs:
Thanks, the revisions look good. I'd say we are ready to merge

rgemulla:
Thanks!

PR #97, opened 2020-05-14
Several improvements for lookup_embedder.penalty

samuelbroscheit:
- improve memory uage for unweighted penalty by using PyTorch builtin norm()

 - use call to embeddings in penalty to make sparse embeddings work, before this fix unewighted penalty for sparse embeddings did not work

### DIFF
--- a/kge/model/embedder/lookup_embedder.py
+++ b/kge/model/embedder/lookup_embedder.py
@@ -29,7 +29,7 @@ class LookupEmbedder(KgeEmbedder):
             self.dim = round_to_points(round_embedder_dim_to, self.dim)
 
         # setup embedder
-        self.embeddings = torch.nn.Embedding(
+        self._embeddings = torch.nn.Embedding(
             self.vocab_size, self.dim, sparse=self.sparse
         )
 
@@ -45,7 +45,7 @@ class LookupEmbedder(KgeEmbedder):
             init_args["a"] = init_args["b"] * -1
             self.set_option("initialize_args.a", init_args["a"], log=True)
 
-        self.initialize(self.embeddings.weight.data, init_, init_args)
+        self.initialize(self._embeddings.weight.data, init_, init_args)
 
         # TODO handling negative dropout because using it with ax searches for now
         dropout = self.get_option("dropout")
@@ -64,37 +64,38 @@ class LookupEmbedder(KgeEmbedder):
 
             def normalize_embeddings(job):
                 if self.normalize_with_grad:
-                    self.embeddings.weight = torch.nn.functional.normalize(
-                        self.embeddings.weight, p=self.normalize_p, dim=-1
+                    self._embeddings.weight = torch.nn.functional.normalize(
+                        self._embeddings.weight, p=self.normalize_p, dim=-1
                     )
                 else:
                     with torch.no_grad():
-                        self.embeddings.weight = torch.nn.Parameter(
+                        self._embeddings.weight = torch.nn.Parameter(
                             torch.nn.functional.normalize(
-                                self.embeddings.weight, p=self.normalize_p, dim=-1
+                                self._embeddings.weight, p=self.normalize_p, dim=-1
                             )
                         )
 
             job.pre_batch_hooks.append(normalize_embeddings)
 
-    def _embed(self, embeddings: Tensor) -> Tensor:
+    def _postprocess(self, embeddings: Tensor) -> Tensor:
         if self.dropout.p > 0:
             embeddings = self.dropout(embeddings)
         return embeddings
 
     def embed(self, indexes: Tensor) -> Tensor:
-        if indexes is None:
-            return self._embed(self._get_all_parameters())
-        else:
-            return self._embed(self.embeddings(indexes.long()))
+        return self._postprocess(self._embeddings(indexes.long()))
 
     def embed_all(self) -> Tensor:
-        return self.embed(None)
-
-    def _get_all_parameters(self) -> Tensor:
-        return self.embeddings(
+        return self.embed(
             torch.arange(
-                self.vocab_size, dtype=torch.long, device=self.embeddings.weight.device
+                self.vocab_size, dtype=torch.long, device=self._embeddings.weight.device
+            )
+        )
+
+    def _embeddings_all(self) -> Tensor:
+        return self._embeddings(
+            torch.arange(
+                self.vocab_size, dtype=torch.long, device=self._embeddings.weight.device
             )
         )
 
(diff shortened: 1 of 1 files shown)


kge/model/embedder/lookup_embedder.py
```diff
@@ -105,13 +107,12 @@ def penalty(self, **kwargs) -> List[Tensor]:
             regularize_weight = self._get_regularize_weight()
             if not self.get_option("regularize_args.weighted"):
                 # unweighted Lp regularization
-                parameters = self.embeddings.weight
```
samuelbroscheit, line 108: This made sparse embeddings break

kge/model/embedder/lookup_embedder.py
```diff
@@ -86,7 +87,8 @@ def embed(self, indexes: Tensor) -> Tensor:
         return self._embed(self.embeddings(indexes.long()))
 
     def embed_all(self) -> Tensor:
-        return self._embed(self.embeddings.weight)
```
samuelbroscheit, line 89: This caused that dropout was incorrectly not applied to score funcs that call embed_all

kge/model/embedder/lookup_embedder.py
```diff
@@ -105,13 +107,12 @@ def penalty(self, **kwargs) -> List[Tensor]:
...
                 # unweighted Lp regularization
-                parameters = self.embeddings.weight
-                if p % 2 == 1:
-                    parameters = torch.abs(parameters)
+                self.all_indexes = self.all_indexes.to(self.embeddings.weight.device)
+                parameters = self.embeddings(self.all_indexes)
                 result += [
                     (
                         f"{self.configuration_key}.L{p}_penalty",
-                        (regularize_weight / p * (parameters ** p)).sum(),
```
samuelbroscheit, line 114: This caused huge memory consumption for RESCAL

kge/model/embedder/lookup_embedder.py
```diff
@@ -86,7 +87,8 @@ def embed(self, indexes: Tensor) -> Tensor:
         return self._embed(self.embeddings(indexes.long()))
 
     def embed_all(self) -> Tensor:
-        return self._embed(self.embeddings.weight)
+        self.all_indexes = self.all_indexes.to(self.embeddings.weight.device)
+        return self.embed(self.all_indexes)
```
rgemulla, line 91: I think this would be handled much better (computation, storage) if a `None` argument was interpreted as "all_indexes" my `embed`.

kge/model/embedder/lookup_embedder.py
```diff
@@ -105,13 +107,12 @@ def penalty(self, **kwargs) -> List[Tensor]:
             regularize_weight = self._get_regularize_weight()
             if not self.get_option("regularize_args.weighted"):
                 # unweighted Lp regularization
-                parameters = self.embeddings.weight
-                if p % 2 == 1:
-                    parameters = torch.abs(parameters)
+                self.all_indexes = self.all_indexes.to(self.embeddings.weight.device)
+                parameters = self.embeddings(self.all_indexes)
```
rgemulla, line 111: Likewise.

samuelbroscheit: Here we do not call embed

kge/model/embedder/lookup_embedder.py
samuelbroscheit: I don't understand how this would improve computation and storage, but I can shift it over to embed. 

so basically you propose:

```
    def embed(self, indexes: Tensor) -> Tensor:
        if indexes is None:
            self.all_indexes = self.all_indexes.to(self.embeddings.weight.device)
            indexes = self.all_indexes 
        return self._embed(self.embeddings(indexes.long()))

```

kge/model/embedder/lookup_embedder.py
rgemulla: but embeddings

kge/model/embedder/lookup_embedder.py
rgemulla: No. Never use all_indexes (do not even create it). Change the API such that indexes=None means all indexes. In this case, we should return self.embeddings.weight. Or am I missing something? It's a waste of space to store a vector of 1...num_embeddings on the GPU.

samuelbroscheit: Accessing self.embeddings.weight was the problem that is being fixed here. The point of the change is that we don't get sparse gradients when we directly access self.embeddings.weight. Please note that sparse gradients is not achieved by self.embeddings.weight.to_sparse(), only the gradients are sparse. 

However, this is fixing the case of using SparseAdam+unweighted penalty, SparseAdam+1vsAll or SparseAdam+KvsAll. One could argue that those are non-sensical anyway because sparse is most likely used when we do not want to access the whole entity embedding table, i.e., for sampled training strategies and weighted penalties. We could just opt to not support that combination and not waste those 7MB / 1M entities of long indices.

rgemulla: I think we should either use `.weight` or `.embeddings()` (i.e., forward), but not mix both. 

I am fine to consistently use `.embeddings()`. But then write `torch.arange(..., device=..., dtype=torch.long)` to get all indexes. I don't think there is any benefit of caching it.

Agreed?

samuelbroscheit: Ok, I thought allocating the exact same tensor constantly is more wasteful, also, with having the GPU-memory fragmentation issue in mind ... . But I'll trust your judgment.

rgemulla: Let's keep it simple and only add caching etc. when we run into problems.

kge/model/embedder/lookup_embedder.py
```diff
@@ -84,11 +83,20 @@ def _embed(self, embeddings: Tensor) -> Tensor:
...
+            return self._embed(self._get_all_parameters())
+        else:
+            return self._embed(self.embeddings(indexes.long()))
 
     def embed_all(self) -> Tensor:
-        self.all_indexes = self.all_indexes.to(self.embeddings.weight.device)
-        return self.embed(self.all_indexes)
+        return self.embed(None)
+
+    def _get_all_parameters(self) -> Tensor:
```
rgemulla, line 94: Names are quite confusing (already before this PR). I suggest:
- Rename `self.embeddings` to `self._embeddings`
- Rename this function to `self._embeddings_all` (for consistency with embed and embed_all)
- Rename `self._embed` to `self._postprocess`

rgemulla:
Thanks!

Issue #2, opened 2019-04-02
Rewrite entity ranking evaluation to sample sp/po pairs and add loss


samuelbroscheit:
https://github.com/rufex2001/kge/blob/9d83e43f5085e4a0d30d70536e4c1772389907cd/kge/job/entity_ranking.py#L72

PR #99, opened 2020-05-16
Add loss/cost on validation data https://github.com/uma-pi1/kge/issues/2

samuelbroscheit:
As title says, this PR adds loss/cost on validation data to address https://github.com/uma-pi1/kge/issues/2 . Reuse the relevant parts of TrainingJob to run the train job on validation data. This also brings the additional feature that TrainJobs can now run on arbitrary splits; all dependencies on the train split are removed and only determined by the necessary arguments that are propagated down.

### DIFF
--- a/kge/job/train.py
+++ b/kge/job/train.py
@@ -9,6 +9,7 @@ from dataclasses import dataclass
 import torch
 import torch.utils.data
 import numpy as np
+from torch.utils.data import DataLoader
 
 from kge import Config, Dataset
 from kge.job import Job
@@ -63,12 +64,12 @@ class TrainingJob(Job):
         self.trace_batch: bool = self.config.get("train.trace_level") == "batch"
         self.epoch: int = 0
         self.valid_trace: List[Dict[str, Any]] = []
-        self.is_prepared = False
+        self.is_prepared: Dict[str, bool] = defaultdict(lambda: False)
         self.model.train()
 
         # attributes filled in by implementing classes
-        self.loader = None
-        self.num_examples = None
+        self.loader: Dict[str, DataLoader] = dict()
+        self.num_examples: Dict[str, int] = dict()
         self.type_str: Optional[str] = None
 
         #: Hooks run after training for an epoch.
@@ -166,7 +167,9 @@ class TrainingJob(Job):
             # start a new epoch
             self.epoch += 1
             self.config.log("Starting epoch {}...".format(self.epoch))
-            trace_entry = self.run_epoch()
+            trace_entry = self.run_epoch(
+                split=self.train_split, echo_trace=True, do_backward=True
+            )
             for f in self.post_epoch_hooks:
                 f(self, trace_entry)
             self.config.log("Finished epoch {}.".format(self.epoch))
@@ -287,14 +290,16 @@ class TrainingJob(Job):
         else:
             self.config.log("No checkpoint found, starting from scratch...")
 
-    def run_epoch(self) -> Dict[str, Any]:
+    def run_epoch(
+        self, split: str, echo_trace: bool, do_backward: bool
+    ) -> Dict[str, Any]:
         "Runs an epoch and returns a trace entry."
 
         # prepare the job is not done already
-        if not self.is_prepared:
-            self._prepare()
+        if not self.is_prepared[split]:
+            self._prepare(split=split)
             self.model.prepare_job(self)  # let the model add some hooks
-            self.is_prepared = True
+            self.is_prepared[split] = True
 
         # variables that record various statitics
         sum_loss = 0.0
@@ -307,14 +312,14 @@ class TrainingJob(Job):
         optimizer_time = 0.0
 
         # process each batch
-        for batch_index, batch in enumerate(self.loader):
+        for batch_index, batch in enumerate(self.loader[split]):
             for f in self.pre_batch_hooks:
                 f(self)
 
             # process batch (preprocessing + forward pass + backward pass on loss)
             self.optimizer.zero_grad()
             batch_result: TrainingJob._ProcessBatchResult = self._process_batch(
-                batch_index, batch
+                batch_index, batch, do_backward
             )
             sum_loss += batch_result.avg_loss * batch_result.size
 
@@ -323,7 +328,7 @@ class TrainingJob(Job):
             penalties_torch = self.model.penalty(
                 epoch=self.epoch,
                 batch_index=batch_index,
-                num_batches=len(self.loader),
+                num_batches=len(self.loader[split]),
                 batch=batch,
             )
             batch_forward_time += time.time()
@@ -332,7 +337,8 @@ class TrainingJob(Job):
             batch_backward_time = batch_result.backward_time - time.time()
             penalty = 0.0
             for index, (penalty_key, penalty_value_torch) in enumerate(penalties_torch):
-                penalty_value_torch.backward()
+                if do_backward:
+                    penalty_value_torch.backward()
                 penalty += penalty_value_torch.item()
                 sum_penalties[penalty_key] += penalty_value_torch.item()
             sum_penalty += penalty
@@ -372,9 +378,11 @@ class TrainingJob(Job):
                     )
 
             # update parameters
-            batch_optimizer_time = -time.time()
-            self.optimizer.step()
-            batch_optimizer_time += time.time()
+            batch_optimizer_time = 0
+            if do_backward:
+                batch_optimizer_time = -time.time()
+                self.optimizer.step()
+                batch_optimizer_time += time.time()
 
             # tracing/logging
             if self.trace_batch:
@@ -382,10 +390,10 @@ class TrainingJob(Job):
                     "type": self.type_str,
                     "scope": "batch",
                     "epoch": self.epoch,
-                    "split": self.train_split,
+                    "split": split,
                     "batch": batch_index,
                     "size": batch_result.size,
-                    "batches": len(self.loader),
+                    "batches": len(self.loader[split]),
                     "lr": [group["lr"] for group in self.optimizer.param_groups],
                     "avg_loss": batch_result.avg_loss,
                     "penalties": [p.item() for k, p in penalties_torch],
@@ -399,29 +407,30 @@ class TrainingJob(Job):
                 for f in self.post_batch_trace_hooks:
                     f(self, batch_trace)
                 self.trace(**batch_trace, event="batch_completed")
-            print(
-                (
-                    "\r"  # go back
-                    + "{}  batch{: "
-                    + str(1 + int(math.ceil(math.log10(len(self.loader)))))
-                    + "d}/{}"
-                    + ", avg_loss {:.4E}, penalty {:.4E}, cost {:.4E}, time {:6.2f}s"
-                    + "\033[K"  # clear to right
-                ).format(
-                    self.config.log_prefix,
-                    batch_index,
-                    len(self.loader) - 1,
-                    batch_result.avg_loss,
-                    penalty,
-                    cost_value,
-                    batch_result.prepare_time
-                    + batch_forward_time
-                    + batch_backward_time
-                    + batch_optimizer_time,
-                ),
-                end="",
-                flush=True,
-            )
+            if echo_trace:
+                print(
+                    (
+                        "\r"  # go back
+                        + "{}  batch{: "
+                        + str(1 + int(math.ceil(math.log10(len(self.loader[split])))))
+                        + "d}/{}"
+                        + ", avg_loss {:.4E}, penalty {:.4E}, cost {:.4E}, time {:6.2f}s"
+                        + "\033[K"  # clear to right
+                    ).format(
+                        self.config.log_prefix,
+                        batch_index,
+                        len(self.loader[split]) - 1,
+                        batch_result.avg_loss,
+                        penalty,
+                        cost_value,
+                        batch_result.prepare_time
+                        + batch_forward_time
+                        + batch_backward_time
+                        + batch_optimizer_time,
+                    ),
+                    end="",
+                    flush=True,
+                )
 
             # update times
             prepare_time += batch_result.prepare_time
@@ -440,14 +449,15 @@ class TrainingJob(Job):
             type=self.type_str,
             scope="epoch",
             epoch=self.epoch,
-            split=self.train_split,
-            batches=len(self.loader),
-            size=self.num_examples,
+            split=split,
+            batches=len(self.loader[split]),
+            size=self.num_examples[split],
             lr=[group["lr"] for group in self.optimizer.param_groups],
-            avg_loss=sum_loss / self.num_examples,
-            avg_penalty=sum_penalty / len(self.loader),
-            avg_penalties={k: p / len(self.loader) for k, p in sum_penalties.items()},
-            avg_cost=sum_loss / self.num_examples + sum_penalty / len(self.loader),
+            avg_loss=sum_loss / self.num_examples[split],
+            avg_penalty=sum_penalty / len(self.loader[split]),
+            avg_penalties={k: p / len(self.loader[split]) for k, p in sum_penalties.items()},
+            avg_cost=sum_loss / self.num_examples[split]
+            + sum_penalty / len(self.loader[split]),
             epoch_time=epoch_time,
             prepare_time=prepare_time,
             forward_time=forward_time,
@@ -458,10 +468,12 @@ class TrainingJob(Job):
         )
         for f in self.post_epoch_trace_hooks:
             f(self, trace_entry)
-        trace_entry = self.trace(**trace_entry, echo=True, echo_prefix="  ", log=True)
+        trace_entry = self.trace(
+            **trace_entry, echo=echo_trace, echo_prefix="  ", log=True
+        )
         return trace_entry
 
-    def _prepare(self):
+    def _prepare(self, split: str):
         """Prepare this job for running.
 
         Sets (at least) the `loader`, `num_examples`, and `type_str` attributes of this
@@ -484,7 +496,7 @@ class TrainingJob(Job):
         backward_time: float
 
     def _process_batch(
-        self, batch_index: int, batch
+        self, batch_index: int, batch, do_backward: bool
     ) -> "TrainingJob._ProcessBatchResult":
         "Run forward and backward pass on batch and return results."
         raise NotImplementedError
@@ -538,23 +550,6 @@ class TrainingJobKvsAll(TrainingJob):
                     )
                 )
 
-        config.log("Initializing 1-to-N training job...")
-        self.type_str = "KvsAll"
-
-        if self.__class__ == TrainingJobKvsAll:
-            for f in Job.job_created_hooks:
-                f(self)
-
-    def _prepare(self):
-        from kge.indexing import index_KvsAll_to_torch
-
-        # determine enabled query types
-        self.query_types = [
-            key
-            for key, enabled in self.config.get("KvsAll.query_types").items()
-            if enabled
-        ]
-
         #' for each query type: list of queries
         self.queries = {}
 
@@ -568,31 +563,48 @@ class TrainingJobKvsAll(TrainingJob):
 
         #' for each query type (ordered as in self.query_types), index right after last
         #' example of that type in the list of all examples
-        self.query_end_index = []
+        self.query_end_index = {}
+
+        config.log("Initializing 1-to-N training job...")
+        self.type_str = "KvsAll"
+
+        if self.__class__ == TrainingJobKvsAll:
+            for f in Job.job_created_hooks:
+                f(self)
+
+    def _prepare(self, split: str):
+        from kge.indexing import index_KvsAll_to_torch
+
+        # determine enabled query types
+        self.query_types = [
+            key
+            for key, enabled in self.config.get("KvsAll.query_types").items()
+            if enabled
+        ]
 
         # construct relevant data structures
-        self.num_examples = 0
+        self.num_examples[split] = 0
         for query_type in self.query_types:
             index_type = (
                 "sp_to_o"
                 if query_type == "sp_"
                 else ("so_to_p" if query_type == "s_o" else "po_to_s")
             )
-            index = self.dataset.index(f"{self.train_split}_{index_type}")
-            self.num_examples += len(index)
-            self.query_end_index.append(self.num_examples)
+            index = self.dataset.index(f"{split}_{index_type}")
+            self.num_examples[split] += len(index)
+            self.query_end_index[query_type + split] = self.num_examples[split]
 
             # Convert indexes to pytorch tensors (as described above).
             (
-                self.queries[query_type],
-                self.labels[query_type],
-                self.label_offsets[query_type],
+                self.queries[query_type + split],
+                self.labels[query_type + split],
+                self.label_offsets[query_type + split],
             ) = index_KvsAll_to_torch(index)
 
         # create dataloader
-        self.loader = torch.utils.data.DataLoader(
-            range(self.num_examples),
-            collate_fn=self._get_collate_fun(),
+        self.loader[split] = torch.utils.data.DataLoader(
+            range(self.num_examples[split]),
+            collate_fn=self._get_collate_fun(split),
             shuffle=True,
             batch_size=self.batch_size,
             num_workers=self.config.get("train.num_workers"),
@@ -602,7 +614,7 @@ class TrainingJobKvsAll(TrainingJob):
             pin_memory=self.config.get("train.pin_memory"),
         )
 
-    def _get_collate_fun(self):
+    def _get_collate_fun(self, split: str):
         # create the collate function
         def collate(batch):
             """For a batch of size n, returns a dictionary of:
@@ -621,11 +633,15 @@ class TrainingJobKvsAll(TrainingJob):
             for example_index in batch:
                 start = 0
                 for query_type_index, query_type in enumerate(self.query_types):
-                    end = self.query_end_index[query_type_index]
+                    end = self.query_end_index[query_type + split]
                     if example_index < end:
                         example_index -= start
-                        num_ones += self.label_offsets[query_type][example_index + 1]
-                        num_ones -= self.label_offsets[query_type][example_index]
+                        num_ones += self.label_offsets[query_type + split][
+                            example_index + 1
+                        ]
+                        num_ones -= self.label_offsets[query_type + split][
+                            example_index
+                        ]
                         break
                     start = end
 
@@ -638,13 +654,13 @@ class TrainingJobKvsAll(TrainingJob):
             for batch_index, example_index in enumerate(batch):
                 start = 0
                 for query_type_index, query_type in enumerate(self.query_types):
-                    end = self.query_end_index[query_type_index]
+                    end = self.query_end_index[query_type + split]
                     if example_index < end:
                         example_index -= start
                         query_type_indexes_batch[batch_index] = query_type_index
-                        queries = self.queries[query_type]
-                        label_offsets = self.label_offsets[query_type]
-                        labels = self.labels[query_type]
+                        queries = self.queries[query_type + split]
+                        label_offsets = self.label_offsets[query_type + split]
+                        labels = self.labels[query_type + split]
                         if query_type == "sp_":
                             query_col_1, query_col_2, target_col = S, P, O
                         elif query_type == "s_o":
@@ -685,7 +701,9 @@ class TrainingJobKvsAll(TrainingJob):
 
         return collate
 
-    def _process_batch(self, batch_index, batch) -> TrainingJob._ProcessBatchResult:
+    def _process_batch(
+        self, batch_index, batch, do_backward: bool
+    ) -> TrainingJob._ProcessBatchResult:
         # prepare
         prepare_time = -time.time()
         queries_batch = batch["queries"].to(self.device)
@@ -756,7 +774,8 @@ class TrainingJobKvsAll(TrainingJob):
                 loss_value_total = loss_value.item()
                 forward_time += time.time()
                 backward_time -= time.time()
-                loss_value.backward()
+                if do_backward:
+                    loss_value.backward()
                 backward_time += time.time()
 
         # all done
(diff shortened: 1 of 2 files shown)


rgemulla:
Why is this needed? 

We already have the capability to run a training job on arbitrary splits and to select the splits for filtering (in case of negative sampling). E.g., one may simply run a training job using "train.split=valid". 

What may be missing is (1) disable the optimizer, (2) just run one epoch, and (3) disable checkpoints. Is this what this PR is doing?

samuelbroscheit:
> We already have the capability to run a training job on arbitrary splits and to select the splits for filtering (in case of negative sampling). E.g., one may simply run a training job using "train.split=valid".
> 
> What may be missing is (1) disable the optimizer, (2) just run one epoch, and (3) disable checkpoints. Is this what this PR is doing?

Yes, there are so many things that have to be disabled (disable optimzer, disable backward, no valid job instantiation, everything in .run() basically, ...) just to run it as a new instance of TrainingJob. I found it to be more sensible to call run_epoch from the valid_job, that has access to its parent_job anyhow. It works very nicely and, of couse, I tested it that it gives the same results as before for all TrainingJobs.

rgemulla:
I see & sounds good. Can you very briefly outline the approach/changes? That simplifies the code review.

rgemulla:
I do think that we should create a new instance of the training job from the valid job though (so that it can have a different config). That's cleaner and more future-proof than changing code in the training jobs.

rgemulla:
This would also allow its use from an arbitrary eval job (not necessarily valid)

samuelbroscheit:
> This would also allow its use from an arbitrary eval job (not necessarily valid)

I agree. However, I started with an approach where the TrainingJob was acting depending on a flag in but that made it so ugly  and hard to read with "if"s everywhere. So I ended up with the current solution. I think about if there is another way, because all we care about is run_epoch(). In the current approach it has the signature

```
trace_entry = self.run_epoch(
                split=self.train_split, echo_trace=True, do_backward=True
            )

```

rgemulla:
That's fine, but you can still create a new job during valid. Then just call `run_epoch`. The first two arguments are not needed (1st = part of job, 2nd = tresult of run_epoch). I'd rename the third argument to `forward_only` .

samuelbroscheit:
Should the training job just produce a seperate trace entry and echo it as well? I think its nicer when we retrieve the trace result and  do

```
            metrics.update(
                avg_loss=train_trace_entry["avg_loss"],
                avg_penalty=train_trace_entry["avg_penalty"],
                avg_penalties=train_trace_entry["avg_penalties"],
                avg_cost=train_trace_entry["avg_cost"],
            )

```

WDYT?

rgemulla:
You mean the evaluation job? I'd only copy the avg_loss, the other entries are depending on config settings and do not actually evaluate the model. They can still be found in the trace entry of the subtraining job if needed.

samuelbroscheit:
> You mean the evaluation job? I'd only copy the avg_loss, the other entries are depending on config settings and do not actually evaluate the model. They can still be found in the trace entry of the subtraining job if needed.

Ok. Should the subtraining job echo its trace as well?

samuelbroscheit:
My current plan is:

create 
`__init_for_train__` and 
`__init_for_eval__`  

which are branched to in `__init__` and 

`run_for_train` and 
`run_for_eval` 

which are branched into from `run()`. The `run_epoch()` trace can be retrieved with a hook.

WDYT

rgemulla:
Is this all needed? Why not just stick with `run_epoch` with flag `forward_only`? 

To ease things up, we could additionally create an EvalTrainingLossJob: it creates the TrainingJob, calls run_epoch, and does everything else that's needed.

samuelbroscheit:
> Is this all needed? Why not just stick with `run_epoch` with flag `forward_only`?

Ah, did you propose to create the TrainJob but not call run() on it, but only run_epoch? Violates the Job-API a bit, but sure that works. 
 
> To ease things up, we could additionally create an EvalTrainingLossJob: it creates the TrainingJob, calls run_epoch, and does everything else that's needed.

 What would be the advantage of that?

rgemulla:
Advantage: (i) that additional job has our Job API and gathers all the "wrapper code" for reuse in other places. E.g., (ii) it could also be run directly from the config (if we wanted that) and (iii) it can be easily used as parts of other jobs (again, if we wanted that). 

For now, key advantage would be (i), not sure if (ii)+(iii) will ever matter.

I am also OK with just using run_epoch for now. It's just that the wrapper job is cleaner, easier to maintain, and probably little additional effort to create.

samuelbroscheit:
Alright makes sense and should be easy to do.T


Rainer Gemulla <notifications@github.com> schrieb am Mi., 20. Mai 2020,
14:41:

samuelbroscheit:
I wonder if eval.type should not  take a list of eval types, runs all of them, and collects their metrics into one trace.

rgemulla:
But this runs multiple eval jobs, right? This could be implemented by having an "aggregate eval" job, which does this. Let's keep this for the future.

samuelbroscheit:
Can you review again?

kge/job/train.py
```diff
@@ -538,6 +547,21 @@ def __init__(self, config, dataset, parent_job=None):
...
+        self.labels = {}
+
+        #' for each query type: list of starting offset of labels in self.labels. The
+        #' labels for the i-th query of query_type are in labels[query_type] in range
+        #' label_offsets[query_type][i]:label_offsets[query_type][i+1]
+        self.label_offsets = {}
+
+        #' for each query type (ordered as in self.query_types), index right after last
+        #' example of that type in the list of all examples
+        self.query_end_index = {}
```
rgemulla, line 563: Why was this moved?

kge/job/train.py
```diff
@@ -555,21 +579,6 @@ def _prepare(self):
...
-        #' label_offsets[query_type][i]:label_offsets[query_type][i+1]
-        self.label_offsets = {}
-
-        #' for each query type (ordered as in self.query_types), index right after last
-        #' example of that type in the list of all examples
-        self.query_end_index = []
-
         # construct relevant data structures
         self.num_examples = 0
         for query_type in self.query_types:
```
rgemulla, line 584: Bit not those parts that fill it? All should stay in prepare

kge/job/train.py
```diff
@@ -756,7 +767,8 @@ def _process_batch(self, batch_index, batch) -> TrainingJob._ProcessBatchResult:
                 loss_value_total = loss_value.item()
                 forward_time += time.time()
                 backward_time -= time.time()
-                loss_value.backward()
+                if forward_only:
```
rgemulla, line 770: if not forward_only (and everywhere else too)

kge/job/train.py
```diff
@@ -812,8 +820,6 @@ def _prepare(self):
             pin_memory=self.config.get("train.pin_memory"),
         )
 
-        self.is_prepared = True
```
rgemulla, line 815: Why is anything changed in prepare?

samuelbroscheit: Those were redundant, is already set in run()

kge/job/train.py
samuelbroscheit: Just for consistency. Usually in our codebase we always declare class instance variables in the init(). I can revert if you want.

rgemulla: I see. I suggest to move back for readability. Alternatively, set to `None` here to make clear that this is not correctly initialized.

kge/job/eval.py
```diff
@@ -30,6 +33,12 @@ def __init__(self, config, dataset, parent_job, model):
         self.filter_with_test = config.get("eval.filter_with_test")
         self.epoch = -1
 
+        train_job_on_eval_split_config = config.clone()
+        train_job_on_eval_split_config.set("train.split", self.eval_split)
+        self.eval_train_loss_job = TrainingJob.create(
+            config=train_job_on_eval_split_config, parent_job=self, dataset=dataset
+        )
```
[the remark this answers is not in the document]
samuelbroscheit: So instanciate `EvalTrainingLossJob` here? Then `EvalTrainingLossJob.run()` returns the trace and also needs to take a "verbose" argument somewhere. Should we add a `verbose` argument to `Job` which defaults to `True`?

kge/job/train.py
rgemulla: Ok

kge/job/eval.py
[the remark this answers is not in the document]
rgemulla: Yes, instantiate the job. 

Adding verbose to job is a good idea, but that requires even more changes (e.g., to apply it consistently) and is better suited to another PR. 

For now, I suggest to use a default-true `verbose` argument only for evaltraininglossjob.

kge/job/eval.py
```diff
@@ -98,6 +111,72 @@ def resume(self, checkpoint_file=None):
...
+            split=self.eval_split,
+            epoch=self.epoch,
+            epoch_time=epoch_time,
+            event="eval_completed",
+            avg_loss=train_trace_entry["avg_loss"],
+        )
+        for f in self.post_epoch_trace_hooks:
+            f(self, trace_entry)
+
+        # if validation metric is not present, try to compute it
```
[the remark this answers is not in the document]
samuelbroscheit: If EvalTrainingLossJob is run standalone and the avg_loss should be postprocessed and used as a metric then it needs to be here. But as this is a general thing that applies to all eval jobs I made it a post_process hook that is run by all eval jobs, i.e., it didn't belong into only entity ranking as well.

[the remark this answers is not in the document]
samuelbroscheit: See commit message.

samuelbroscheit:
I adressed your reviews. 

I ended up doing a small revision of EvaluationJob: EvaluationJob.run() is implemented now in EvaluationJob which does standard stuff that has to be done for every EvaluationJob and then calls self._run() (suggestions for better name for self.._run() ?).

I shifted EntityRanking specific stuff from EvaluationJob to there.

I added self.verbose to EvaluationJob and use it. 

I made EvalTrainingLossJob an instance variable of EvaluationJob and run it as post_epoch_trace_hook.  This can be made nicer when we create AggregatingEvalJob that runs and aggregates a list of EvaluationJobs. Then the post_epoch_trace_hook can be removed from EvaluationJob and can be replaced by the entry "eval.type: entity_ranking, training_loss" in config-default-yaml. See https://github.com/uma-pi1/kge/issues/102

kge/job/train.py
```diff
@@ -51,21 +52,23 @@ def __init__(
...
-            valid_conf, dataset, parent_job=self, model=self.model
-        )
         self.config.check("train.trace_level", ["batch", "epoch"])
         self.trace_batch: bool = self.config.get("train.trace_level") == "batch"
         self.epoch: int = 0
-        self.valid_trace: List[Dict[str, Any]] = []
         self.is_prepared = False
         self.model.train()
 
+        if config.get("job.type") == "train":
```
rgemulla, line 61: I suggest to disable validation with an optional constructor argument instead of this more hacky approach.

kge/job/eval.py
```diff
@@ -98,6 +166,50 @@ def resume(self, checkpoint_file=None):
         )
 
 
+class EvalTrainingLossJob(EvaluationJob):
```
rgemulla, line 169: Why not TrainingLossEvaluationJob for consistency?

kge/job/train.py
```diff
@@ -621,7 +635,7 @@ def collate(batch):
             for example_index in batch:
                 start = 0
                 for query_type_index, query_type in enumerate(self.query_types):
```
rgemulla, line 637: drop enumerate if not needed anymore

kge/job/train.py
```diff
@@ -638,7 +652,7 @@ def collate(batch):
             for batch_index, example_index in enumerate(batch):
                 start = 0
                 for query_type_index, query_type in enumerate(self.query_types):
```
rgemulla, line 654: likewise

rgemulla:
Please also add a CHANGELOG entry before you merge this PR.

kge/job/train.py
samuelbroscheit: Only as constructor or as a config?

rgemulla: In the config, we have it already. That being said, forget the constructor argument and simply set "valid.every" to 0 in the TrainingLossEvacluatinJob.

rgemulla: Likewise, set train.checkpoint.every to 0.

samuelbroscheit: Hm, I just realized that I was still initializing optimizer (and loss). Optimizer creates new parameters based on model.parameters(). In this context querying job.type makes more sense. See new commit.

kge/job/eval.py
```diff
@@ -175,13 +175,14 @@ def __init__(self, config: Config, dataset: Dataset, parent_job, model):
 
         train_job_on_eval_split_config = config.clone()
         train_job_on_eval_split_config.set("train.split", self.eval_split)
+        train_job_on_eval_split_config.set("negative_sampling.filtering.split", self.config.get("train.split"))
```
rgemulla, line 178: Only do this if unset.

kge/job/train.py
rgemulla: You are assuming that the outside eval job does not have job.type=train (it could have). Either set this outside or add a constructor argument "initialize_for_forward_only=False" (preferred since clearer, I think).

kge/util/sampler.py
```diff
@@ -148,7 +148,12 @@ def _filter_and_resample(
         cols = [[P, O], [S, O], [S, P]][slot]
         pairs = positive_triples[:, cols]
         for i in range(positive_triples.size(0)):
-            positives = index.get((pairs[i][0].item(), pairs[i][1].item())).numpy()
+            pair = (pairs[i][0].item(), pairs[i][1].item())
+            positives = (
+                index.get(pair).numpy()
+                if pair in index
+                else torch.IntTensor([]).numpy()
```
rgemulla, line 155: This tensor should contain the positive and not be empty

rgemulla: Also, why not create a numpy array right away?

kge/util/sampler.py
```diff
@@ -277,13 +280,18 @@ def _filter_and_resample_fast(
         positives_index = numba.typed.Dict()
         for i in range(batch_size):
             pair = (pairs[i][0], pairs[i][1])
-            positives_index[pair] = index.get(pair).numpy()
+            positives_index[pair] = (
+                index.get(pair).numpy()
+                if pair in index
+                else torch.IntTensor([]).numpy()
```
rgemulla, line 286: likewise here

rgemulla:
Your last changes raise the question of whether all training jobs handle indexes correctly when used in an "eval only" setting. E.g., for KvsAll, what are the labels being used? Training data? Training + eval data?

kge/job/train.py
samuelbroscheit: > You are assuming that the outside eval job does not have job.type=train (it could have). 

Huh? Isn't it one config == one job? "job.type" is either eval, train, search ..... When I pass trough the eval's config, then its job.type is "eval". I must misunderstand what you are saying here?

> Either set this outside or add a constructor argument "initialize_for_forward_only=False" (preferred since clearer, I think).

Ok can do.

samuelbroscheit:
> Your last changes raise the question of whether all training jobs handle indexes correctly when used in an "eval only" setting. E.g., for KvsAll, what are the labels being used? Training data? Training + eval data?

True, for eval only KvsAll should use Training + eval data.

kge/job/train.py
rgemulla: Not if the job is created via EvaluationJob.create(...). An explicit argument makes this clearer

rgemulla:
For the other jobs it's clearer, I guess, but I feel taht loss can turn out to be misleading.

I think we should not make the training_loss a default for the evaluation job (i.e., do not add it as a hook). It may be too confusing for a default since it depends on training/setup so much.

samuelbroscheit:
> For the other jobs it's clearer, I guess, but I feel taht loss can turn out to be misleading.
> 
> I think we should not make the training_loss a default for the evaluation job (i.e., do not add it as a hook). It may be too confusing for a default since it depends on training/setup so much.

Actually also for KvsAll its clear, its the same as for all losses. It should only be the labels from eval data, i.e. for unseen labels. 

The only particular thing with NS is, that the current implementation does the filtering of positive triples based on what it is given as `train.split`. But actually that is always the known KG, i.e., the train split.

rgemulla:
Why is it clear for KvsAll? If done like you state, the losses are not comparable between train and valid (valid loss will be pretty much like 1vsAll because the data probably looks like this). Nevertheless this may be the best option. Another sensible choice may be the prefixes/suffixes from eval, but the labels from train+eval.

rgemulla:
For negative sampling, the filtering split (or splits?) can already be specified, I think.

samuelbroscheit:
> For negative sampling, the filtering split (or splits?) can already be specified, I think.

Yes but at the moment it is automagically the split defined as `train.split`. However, `train.split` is set to "valid" when TrainingJob is run as evaluation. https://github.com/uma-pi1/kge/blob/f2bcbdb47776e6a3b73d07cca28603bbce723f88/kge/util/sampler.py#L33-L34

rgemulla:
> > For negative sampling, the filtering split (or splits?) can already be specified, I think.
> 
> Yes but at the moment it is automagically the split defined as `train.split` if not set otherwise.
So which one to pick here? The most natural choice would be train+valid, but that does not seem to be supported.

samuelbroscheit:
> Why is it clear for KvsAll? If done like you state, the losses are not comparable between train and valid (valid loss will be pretty much like 1vsAll because the data probably looks like this). Nevertheless this may be the best option. Another sensible choice may be the prefixes/suffixes from eval, but the labels from train+eval.

When we were talking about train+eval, this is what I meant, prefixes from valid, but labels from both. I understand what you mean that the loss on train and the loss on valid are not comparable in that case. With "it is clear" I meant that the only-on-valid KvsAll loss would still work to measure overfitting. Would be nicer if it would be a comparable loss, though.

samuelbroscheit:
> > > For negative sampling, the filtering split (or splits?) can already be specified, I think.
> > 
> > 
> > Yes but at the moment it is automagically the split defined as `train.split` if not set otherwise.
> 
> So which one to pick here? The most natural choice would be train+valid, but that does not seem to be supported.

So should we?

rgemulla:
If we allow filtering with train+valid, we'd need to change the interface. Not sure if we want to go there; it significantly complicates the code if we allow for multiple splits during training.

An alternative may be to create new splits by merging existing splits. That's wasting memory, but at least it's less intrusive. The easiest way to do this is (1) extend the dataset.yaml with support for multple files per split, (2) add a train_valid and a train_valid_test split with the appropriate files. That should be a separate PR. For now, why not simply use the valid split and issue a warning or an error when KvsAll is used like that?

rgemulla:
Also, using KvsAll with a different "query split" and "label split" is highly confusing. 

Option 1: TrainingLossEvaluationJob solely uses the validation split and ignores all other splits (also for negative sampling etc). This way, it's clear what it does and we can issue a warning saying this. How useful such a loss is, is another question.

Option 2: we do not add this method of evaluation. For comparabiliy with training data, we can proceed the other way around: run the standard eval job on a sample of the training data (this sampling can be done externally to create a new train_sample split).

rgemulla:
I tend to option 2.

samuelbroscheit:
I'm for option 3: do it and do it correctly. 

Now we're almost there and understand the implications. Would be a waste, now that we scrutinized it. I also like that this is tightly coupled to the filtering of entity_ranking. This would mean that EvaluationJob and TrainingJob both have a query_split and a label_splits attribute. label_splits would replace filter_splits in EvaluationJob. It stays an automatically set option and is not a configurable option. We can also call it "known answers". 

For the TrainingJob when not in eval mode -> nothing changes. query split == label split. 

In eval mode:
- For 1vsAll in eval mode: label_split == query_split. 

- For NS in eval mode: label_split only comes into play when positives are filtered -> label_split = train split + eval split(s)

- KvsAll  in eval mode:  label_split are the labels -> label_split = train split + eval split(s)
For KvsAll we do not need extra dataset files. Its easy to merge the dictionaries and remove the keys that are not in train_split.

rgemulla:
That would also be a suitable option. The main drawback is that it's more work. 

I suggest a slightly revised approach:

For 1vsAll: do nothing
For NS: add support for multiple filter splits (not just one, as we have now)
For KvsAll: add a label_splits option (under key KvsAll) along the lines you discussed

For NS: rename negative_sampling.filtering.split to negative_sampling.filtering.splits. Automatically add train.split, but additionally use the splits mentioned here (along the lines we already do it for eval.filter_splits).

For KvsAll: You are right in that the label_splits do not need to be merged apriori (which would mean that we need to copy all data). This can be done in the respective collate() function during data loading. If label_splits is unset, train.split should be used. If set, these splits + train.split should be used (again, along the lines we already do it for eval.filter_splits).

samuelbroscheit:
https://github.com/uma-pi1/kge/pull/99/commits/083b27cd977175d7aa30e7467adf5f1c36055eea contains all the relevant changes

samuelbroscheit:
Can you review https://github.com/uma-pi1/kge/commit/66a9bc17d9232a8172cb151df4d4dbfed9b638f3 or is there anything else?

rgemulla:
I cannot review this PR as such. Instead of rebasing it on master, it contains tons of master commits which make it hard to review. Can you rebase on master so that just the changes are here?

samuelbroscheit:
Now?

rgemulla:
Even worse, you destroyed it. Do not touch, I force push in a minute.

kge/config-default.yaml
```diff
@@ -235,6 +235,10 @@ KvsAll:
     s_o: False
     _po: True
 
+  # Dataset splits from which the labels for a query are taken from. Default: If
+  # nothing is specified, then the train split is used.
+  label_splits: []
```
rgemulla, line 240: Semantics should be: additional splits to train split.

kge/config-default.yaml
```diff
@@ -264,7 +268,7 @@ negative_sampling:
     p: False       # as above
     o: False       # as above
 
-    split: ''      # split containing the positives; default is train.split
+    splits: []     # splits containing the positives; default is train.split
```
rgemulla, line 271: Likewise here.

kge/config-default.yaml
```diff
@@ -327,7 +331,8 @@ eval:
   # mean_reciprocal_rank_filtered_with_test.
   filter_with_test: True
 
-  # Type of evaluation (entity_ranking only at the moment)
+  # Type of evaluation (entity_ranking or training_loss). Currently, 
+  # entity_ranking runs training_loss as well.
```
rgemulla, line 335: I vote for disabling the automatic running of training_loss. Makes evaluation too costly.

kge/indexing.py
```diff
@@ -222,13 +223,31 @@ def _invert_ids(dataset, obj: str):
     dataset.config.log(f"Indexed {len(inv)} {obj} ids", prefix="  ")
 
 
+def merge_KvsAll_indexes(dataset, split, key):
+    value = dict([("sp", "o"), ("po", "s"), ("so", "p")])[key]
+    split_combi_str = "_".join(sorted(split))
+    index_name = f"{split_combi_str}_{key}_to_{value}"
+    indexes = [dataset.index(f"{_split}_{key}_to_{value}") for _split in split]
+    dataset._indexes[index_name] = merge_dicts_of_1dim_torch_tensors(indexes)
+    return dataset._indexes[index_name]
```
rgemulla, line 232: Instead of merging indexes (very expensive memory wise), the code should support using multiple indexes.

kge/indexing.py
```diff
@@ -222,13 +223,31 @@ def _invert_ids(dataset, obj: str):
...
                 index_KvsAll, split=split, key=key
             )
+    # create all combinations of splits of length 2 and 3
+    for split_combi in powerset(dataset.files_of_type("triples"), [2, 3]):
+        for key, value in [("sp", "o"), ("po", "s"), ("so", "p")]:
+            split_combi_str = "_".join(sorted(split_combi))
+            index_name = f"{split_combi_str}_{key}_to_{value}"
+            dataset.index_functions[index_name] = IndexWrapper(
+                merge_KvsAll_indexes, split=split_combi, key=key
+            )
```
rgemulla, line 249: This makes it even worse.

kge/job/eval.py
```diff
@@ -31,9 +25,11 @@ def __init__(self, config, dataset, parent_job, model):
         self.filter_splits = self.config.get("eval.filter_splits")
         if self.eval_split not in self.filter_splits:
             self.filter_splits.append(self.eval_split)
-        self.filter_with_test = config.get("eval.filter_with_test")
         self.epoch = -1
 
+        self.verbose = True
```
rgemulla, line 30: Drop this. NOt needed anymore.

kge/job/eval.py
```diff
@@ -31,9 +25,11 @@ def __init__(self, config, dataset, parent_job, model):
         self.filter_splits = self.config.get("eval.filter_splits")
         if self.eval_split not in self.filter_splits:
             self.filter_splits.append(self.eval_split)
-        self.filter_with_test = config.get("eval.filter_with_test")
```
rgemulla, line 34: accidental delete? where was this moved?

kge/job/eval.py
```diff
@@ -64,6 +60,22 @@ def __init__(self, config, dataset, parent_job, model):
...
+        if not isinstance(self, TrainingLossEvaluationJob):
+            self.eval_train_loss_job = TrainingLossEvaluationJob(
+                config, dataset, parent_job=self, model=model
+            )
+            self.eval_train_loss_job.verbose = False
+            self.post_epoch_trace_hooks.append(
+                lambda job, trace: trace.update(
+                    avg_loss=self.eval_train_loss_job.run()["avg_loss"]
+                )
+            )
```
rgemulla, line 77: drop this.

kge/job/eval.py
```diff
@@ -132,8 +200,65 @@ def create_from(
         return super().create_from(checkpoint, new_config, dataset, parent_job)
 
 
-# HISTOGRAM COMPUTATION ###############################################################
+class TrainingLossEvaluationJob(EvaluationJob):
+    """ Entity ranking evaluation protocol """
+
+    def __init__(self, config: Config, dataset: Dataset, parent_job, model):
+        super().__init__(config, dataset, parent_job, model)
+        self.is_prepared = True
```
rgemulla, line 208: Don't do this. Simply use an empty _prepare method. Perhaps you can even let it emtpy in the supervlass instead of using NotImplementedError

kge/misc.py
```diff
@@ -15,6 +18,33 @@ def is_number(s, number_type):
         return False
 
 
+def merge_dicts_of_1dim_torch_tensors(keys_from_dols, values_from_dols=None):
```
rgemulla, line 21: these methods can be dropped with the changes suggested above

kge/util/sampler.py
```diff
@@ -29,9 +29,9 @@ def __init__(self, config: Config, configuration_key: str, dataset: Dataset):
                 "Without replacement sampling is only supported when "
                 "shared negative sampling is enabled."
             )
-        self.filtering_split = config.get("negative_sampling.filtering.split")
-        if self.filtering_split == "":
-            self.filtering_split = config.get("train.split")
+        self.filtering_splits: List[str] = config.get("negative_sampling.filtering.splits")
+        if len(self.filtering_splits) == 0:
+            self.filtering_splits.append(config.get("train.split"))
```
rgemulla, line 34: always append when not there. also, add handling of depracated configs to config.py

kge/util/sampler.py
```diff
@@ -43,7 +43,10 @@ def __init__(self, config: Config, configuration_key: str, dataset: Dataset):
             # otherwise every worker would create every index again and again
             if self.filter_positives[slot]:
                 pair = ["po", "so", "sp"][slot]
-                dataset.index(f"{self.filtering_split}_{pair}_to_{slot_str}")
+                for filtering_split in self.filtering_splits:
+                    dataset.index(f"{filtering_split}_{pair}_to_{slot_str}")
+                filtering_splits_str = '_'.join(sorted(self.filtering_splits))
+                dataset.index(f"{filtering_splits_str}_{pair}_to_{slot_str}")
```
rgemulla, line 49: drop

kge/util/sampler.py
rgemulla: Still open.

kge/util/sampler.py
```diff
@@ -279,13 +287,18 @@ def _filter_and_resample_fast(
         positives_index = numba.typed.Dict()
         for i in range(batch_size):
             pair = (pairs[i][0], pairs[i][1])
-            positives_index[pair] = index.get(pair).numpy()
+            positives_index[pair] = (
+                index.get(pair).numpy()
+                if pair in index
```
rgemulla, line 292: instead, check here for whether it's in one of the indexes etc

samuelbroscheit:
> I rebased (hopefully correctly) and added some review comments.

No its a bit broken, but its fixable.

rgemulla:
Yes, I did not know your new code well enough at places. I think all the echo_true and verbose things need to go out, but you'll be much quicker with this.

kge/config-default.yaml
samuelbroscheit: Validation and test data are a lot smaller (1/10 is the worst valid/train ratio in our current datasets).

kge/indexing.py
samuelbroscheit: Why? They are not invoked. 

As a sidenote: This was neccasary because there is currently no way to determine the cross-validation semantics of a dataset file (train,valid,test). We should (in another PR) add a xvaltype attribute to the dataset files.

kge/job/eval.py
samuelbroscheit: entity_ranking

kge/misc.py
samuelbroscheit: I still need this method for KvsAll, right?

kge/indexing.py
samuelbroscheit: For NS I see that it works, for KvsAll I have to merge them. But granted, not neccessarily as an index.

kge/indexing.py
rgemulla: There are also not needed. Let's keep the code clean and not add "orphans"

rgemulla: Adding more information to the dataset configuration is a good idea (if useful for something).

kge/indexing.py
rgemulla: What I meant here: the correspoding training jobs should support this. Alternatively, one may write a wrapper index that takes muliple indexes and merges there results on demand.

rgemulla: For KvsAll, you can also just add the labels of each label split sequentially in the collate function.

rgemulla: We should not allow using label_splits other than for forward only. Otherwise they leak valid/test data. Likewise for negative sampling, I guess.

PR #108, opened 2020-05-26
fix loading of model

unmeshvrije:
When loading the pre-trained model, some keys in the dictionary need changes.


rgemulla:
Thanks unmehsvrije. We had also realized this and pushed a fix yesterday (f2bcbdb). Does this fix work for you, too?

rgemulla:
Ohh, I see, we forgot some keys. Thanks a lot!

Issue #96, opened 2020-05-08
Support for different tie-breaking schemes


DbCrWk:
Currently, ties are handled in a fairly straightforward way:

https://github.com/uma-pi1/kge/blob/75c45ec1c86d9f72027f9d8b7e046bcb38f35c19/kge/job/entity_ranking.py#L547

In short, all triples that have the same score are assigned the same rank. This rank is the "average rank" of the set of triples.

However, based on a recent paper, [A Re-evaluation of Knowledge Graph Completion Methods](https://arxiv.org/pdf/1911.03903.pdf) for more information, there seems to be some merit to allowing for either custom, pre-defined, or the RANDOM tie-breaking procedure. At the very least, it might be nice to explicitly clarify the tie-breaking procedure in configuration, since tie-breaking does strongly influence results for certain models.

rgemulla:
Thanks! I am closing #21 (a dup) because this issues here provides more context.

Generally, the ranking we use in evaluation is fair in that it does not lead to the inflated performance numbers that plague some other implementations. 

Personally, I am not fond of "random" evaluation results. A better way would be to report the expected metric; that's what's approximated by RANDOM in the paper you mention and it can be exactly computed for MRR/Hits@k.

As always, PRs are welcome!

DbCrWk:
@rgemulla I agree with you (and as you mention, the current method is precisely the expectation of the RANDOM method).

I should have been more clear in my original post: I'm advocating for adding a tie-breaking (or ranking) configuration option, not necessarily even implementing these options in LibKGE. That way, anyone looking at an assignment configuration file will know how ties are handled and it will be more clear that this is, in fact, being handled in a consistent and fair manner. In other words, this paper points out that tie-breaking procedures seem to have a strong enough effect that it is worth notating explicitly what tie-breaking procedure was used when running an experiment.

rgemulla:
That's a good, quick to do suggestion.

Note that what LigKGE currently does is not equivalent to the expectation of the RANDOM method. This is because computing a metric using expected rank (which we do) and computing the expected metric (which RANDOM approximates) is generally different since the metrics are not linear in the rank.

DbCrWk:
@rgemulla awesome!

I think I may have some misunderstanding about what RANDOM is doing and I'll defer to you in this matter.

rgemulla:
Addressed by #110

DbCrWk:
thanks!

PR #110, opened 2020-05-28
(WIP) Tie-breaking extensibility

Nzteb:
Tie-breaking options/documentation, see #96

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -330,20 +330,20 @@ eval:
   # Type of evaluation (entity_ranking only at the moment)
   type: entity_ranking
 
-  # How to handle cases where the correct query answer has the same score as one or
-  # multiple wrong query answers. For example: Query: (s, p, ?). (Filtered) Answers:
-  # ['a': score 10, 'b': score 10, 'c': score 10, 'd': score 11, 'e': score 9].
-  # Correct query answer: 'a'. Possible options are:
-  # - rounded_mean_rank: Calculate the rounded mean rank. The final rank is
-  #                      calculated as 1 + #scores_higher +
-  #                      round_down(#answers_with_same_score / 2). In the example,
-  #                      the final rank for correct answer (s, p,'a') is
-  #                      1 + 1 + round_down(3 / 2) = 3.
-  # - best_rank:         Assign the lowest rank for the correct answer. The final
-  #                      rank for answer (s, p, 'a') is 2. DO NOT USE THIS OPTION, it
-  #                      leads to misleading results.
-  # - worst_rank:        Assign the highest rank for the correct answer. The final
-  #                      rank for answer (s, p, 'a') is 4.
+  # How to handle cases ties between the correct answer and other answers, e.g.,
+  #  Query: (s, p, ?).
+  #  Answers and score: a:10, b:10, c:10, d:11, e:9
+  #  Correct: 'a'.
+  #
+  # Possible options are:
+  # - worst_rank:        Use the highest rank of all answers that have the same
+  #                      score as the correct answer. In example: 4.
+  # - best_rank:         Use the lowest rank of all answers that have the same
+  #                      score as the correct answer (competition scoring). In
+  #                      example: 2. DO NOT USE THIS OPTION, it leads to
+  #                      misleading evaluation results.
+  # - rounded_mean_rank: Average between worst and best rank, rounded up
+  #                      (rounded fractional ranking). In example: 3.
   tie_handling: rounded_mean_rank
 
   # Compute Hits@K for these choices of K
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -359,6 +359,17 @@ eval:
   # Other options
   pin_memory: False
 
+entity_ranking:
+  # How to handle cases where the correct query answer has the same score as one or
+  # multiple wrong query answers, for example: Query: (s,p,?).  (Filtered) Answers:
+  # ['a': score 10, 'b': score 10, 'c': score 10, 'd': score 11]. Correct query
+  # answer: 'a'. Possible options are:
+  # - rounded_mean_rank: Calculate the rounded mean rank. The final rank is calculated
+  #                      as 1 + #scores_higher + #answers_with_same_score // 2,
+  #                      where "//" denotes the floor-division. In the example, the
+  #                      final rank for correct answer (s,p,'a') is 1 + 1 + 3 // 2 = 3.
+  tie_handling: rounded_mean_rank
+
 
 # Configuration options for model validation/selection during training. Applied
 # in addition to the options set under "eval" above.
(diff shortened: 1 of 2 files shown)


kge/config-default.yaml
```diff
@@ -359,6 +359,21 @@ eval:
   # Other options
   pin_memory: False
 
+entity_ranking:
```
rgemulla, line 362: Move this under eval for now. It's a good idea to have separate keys in general, but also other things would need to be moved (e.g., the filter splits). We can do this later in a separate PR.

kge/config-default.yaml
```diff
@@ -359,6 +359,21 @@ eval:
...
+  # How to handle cases where the correct query answer has the same score as one or
+  # multiple wrong query answers, for example: Query: (s, p, ?).  (Filtered) Answers:
+  # ['a': score 10, 'b': score 10, 'c': score 10, 'd': score 11]. Correct query
```
rgemulla, line 365: The example should have one answer above and one below the 10's.

kge/config-default.yaml
```diff
@@ -359,6 +359,21 @@ eval:
...
+  # answer: 'a'. Possible options are:
+  # - rounded_mean_rank: Calculate the rounded mean rank. The final rank is calculated
+  #                      as 1 + #scores_higher + #answers_with_same_score // 2,
```
rgemulla, line 368: write as round_up(...), use /2. Easier to read this way.

kge/config-default.yaml
```diff
@@ -359,6 +359,21 @@ eval:
...
+  #                      where "//" denotes the floor-division. In the example, the
+  #                      final rank for correct answer (s, p,'a') is 1 + 1 + 3 // 2 = 3.
+  # - top:               Assign the best possible rank for the correct answer. The
+  #                      final rank for answer (s, p, 'a') is 2.
```
rgemulla, line 372: Write: DO NOT USE THIS, it leads to misleading evaluation results.

kge/config-default.yaml
```diff
@@ -359,6 +359,21 @@ eval:
...
+  # - bottom:            Assign the worst possible rank for the correct answer. The
+  #                      final rank for answer (s, p, 'a') is 4.
```
rgemulla, line 374: worst possible - highest. Key name is confusing (bottom for highest rank). Perhaps: worst_rank. LIkewise top -> best_rank.

kge/job/entity_ranking.py
```diff
@@ -44,6 +47,17 @@ def _prepare(self):
             pin_memory=self.config.get("eval.pin_memory"),
         )
 
+        # assign a tie breaker
+        tie_handling = self.config.get("entity_ranking.tie_handling")
```
rgemulla, line 51: store this in self.tie_handling

kge/job/entity_ranking.py
```diff
@@ -44,6 +47,17 @@ def _prepare(self):
...
+        if tie_handling == "rounded_mean_rank":
```
rgemulla, line 52: And move this to get_ranks. Gives us more flexibility later on, if we want to use better strategies (such as mean_metric).

kge/config-default.yaml
Nzteb: I have addressed the review. Only here I used round_down() as this matches X//Y.

kge/config-default.yaml
```diff
@@ -330,6 +330,23 @@ eval:
...
   type: entity_ranking
 
+  entity_ranking:
+      # How to handle cases where the correct query answer has the same score as one or
+      # multiple wrong query answers. For example: Query: (s, p, ?). (Filtered) Answers:
+      # ['a': score 10, 'b': score 10, 'c': score 10, 'd': score 11, 'e': score 9].
+      # Correct query answer: 'a'. Possible options are:
+      # - rounded_mean_rank: Calculate the rounded mean rank. The final rank is
+      #                      calculated as 1 + #scores_higher +
+      #                      round_down(#answers_with_same_score / 2). In the example,
```
rgemulla, line 340: We always round up to err on the worst_rank side.

kge/config-default.yaml
```diff
@@ -330,6 +330,23 @@ eval:
...
+      #                      the final rank for correct answer (s, p,'a') is
+      #                      1 + 1 + round_down(3 / 2) = 3.
+      # - best_rank:         Assign the lowest rank for the correct answer. The final
+      #                      rank for answer (s, p, 'a') is 2. We do not recommend to
+      #                      use this option, it leads to misleading results.
```
rgemulla, line 345: To weak. Write DO NOT USE THIS OPTION. It's the reason for a number of problematic results/papers

kge/job/entity_ranking.py
```diff
@@ -13,6 +13,11 @@ class EntityRankingJob(EvaluationJob):
 
     def __init__(self, config: Config, dataset: Dataset, parent_job, model):
         super().__init__(config, dataset, parent_job, model)
+        self.config.check(
+            "eval.entity_ranking.tie_handling",
```
rgemulla, line 17: Why not: eval.tie_handling. It makes sense for all ranking tasks, even the ones we still plan to implement

Nzteb: Yes that also makes sense. So dropping the key 'entity_ranking' completely for now.

kge/config-default.yaml
rgemulla: Ahh, I read this as "other_answers_with_same_score". Prehaps clarifiy: rounded (up) mean rank,

Nzteb: I'm just describing what the code does.  And in ranks + num_ties // 2 ,  "//" rounds the rank down?

rgemulla: See, you are confused as well. We round the number of ties (>=1) down, which corresponds to rounding the mean rank up. E.g., if there is **one** other answer with he same score than the correct one, then num_ties=**2**. We round down num_ties/2, but this corresponds to rounding up the mean rank. Perhaps its best to not show this computation here, it's only conusing. Mean rank rounded up is clear enough.

kge/config-default.yaml
```diff
@@ -330,6 +330,22 @@ eval:
...
+  # Possible options are:
+  # - worst_rank:        Use the highest rank of all answers that have the same
+  #                      score as the correct answer. In example: 4.
+  # - best_rank:         Use the lowest rank of all answers that have the same
+  #                      score as the correct answer (competition scoring). In
+  #                      example: 2. DO NOT USE THIS OPTION, it leads to
+  #                      misleading evaluation results.
+  # - rounded_mean_rank: Average between worst and best rank, rounded up
+  #                      (rounded fractional ranking). In example: 3.
+  tie_handling: rounded_mean_rank
```
rgemulla, line 347: How about this?

rgemulla:
Thanks!

kge/config-default.yaml
Nzteb: Cool!

PR #112, opened 2020-06-05
Initialize with a packaged model

AdrianKs:
This PR allows to initialize a model with the embeddings stored in a packaged model.
This allows for example to load embeddings trained on fb15k-237 into a fb15k train job.


AdrianKs:
One problem remains: How do we want to handle this on resume? Currently if a packaged model is specified in the config it will also load it on resume.
Should I just not load the pretrained embeddings on resume and log that the packaged model was ignored?

kge/model/kge_model.py
```diff
@@ -73,12 +73,20 @@ def load(self, savepoint):
...
+            "_relation_embedder.embeddings.weight",
+            "_base_model._entity_embedder.embeddings.weight",
+            "_base_model._relation_embedder.embeddings.weight",
+        ]
+        good_keys = [
+            "_entity_embedder._embeddings.weight",
+            "_relation_embedder._embeddings.weight",
+            "_base_model._entity_embedder._embeddings.weight",
+            "_base_model._relation_embedder._embeddings.weight",
+        ]
```
rgemulla, line 87: This would be better if it where a dict with key old_name and value new_name.

kge/model/kge_model.py
```diff
@@ -89,6 +97,9 @@ def load(self, savepoint):
         self.load_state_dict(state_dict)
         self.meta = savepoint[1]
 
+    def init_pretrained(self, packaged_model: Dict):
+        raise NotImplementedError()
```
rgemulla, line 101: shouldn't be here.

kge/model/kge_model.py
```diff
@@ -219,6 +230,9 @@ def __init__(self, config: Config, dataset: Dataset, configuration_key: str):
 
         self.dim: int = self.get_option("dim")
 
+    def init_pretrained(self, packaged_model: Dict, dataset_ids: List):
+        raise NotImplementedError()
```
rgemulla, line 234: Also not here.

kge/model/kge_model.py
```diff
@@ -431,6 +445,33 @@ def create_from(
         model.eval()
         return model
 
+    def init_pretrained(self, packaged_model: Dict):
+        entity_package = {
+            "model": dict(
+                [
+                    (k.split("_entity_embedder.")[1], v)
+                    for k, v in packaged_model["model"][0].items()
+                    if k.startswith("_entity_embedder")
```
rgemulla, line 454: This won't work (see good_keys above). 

Option 1: check whether there is exactly one key that ends with ._entity_embedder.embeddings.weight and use that one.
Option 2: just load the model, the get its entity embedder (cleaner, I think).

kge/dataset.py
```diff
@@ -155,7 +155,7 @@ def save_to(self, checkpoint: Dict, meta_keys: Optional[List[str]] = None) -> Di
         meta_checkpoint = {}
         for key in meta_keys:
             meta_checkpoint[key] = self.map_indexes(None, key)
-        checkpoint["dataset"]["meta"] = dataset_checkpoint
+        checkpoint["dataset"]["meta"] = meta_checkpoint
```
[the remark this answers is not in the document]
AdrianKs: cherry-picked

AdrianKs:
this is ready for another review

kge/model/kge_model.py
AdrianKs: implemented option 2

kge/model/embedder/lookup_embedder.py
```diff
@@ -77,6 +78,48 @@ def normalize_embeddings(job):
 
             job.pre_batch_hooks.append(normalize_embeddings)
 
+    @torch.no_grad()
+    def init_pretrained(
+        self,
+        pretrained_embedder: torch.nn.Embedding,
+        ids_pretrained_model: List,
+        ids_new_model: List,
```
rgemulla, line 86: Are these two parameters really needed? Both this and the pretrained embedder know their dataset, which in turn knows the IDs.

kge/model/embedder/lookup_embedder.py
```diff
@@ -77,6 +78,48 @@ def normalize_embeddings(job):
...
+                                  of the pre-trained embedding layer
+            ids_new_model: entity/relation ids corresponding to the embeddings of the
+                           model to be initialized
+            ensure_all: ensure that all entities/relations can be initialized with
+                        embeddings provided in the pre-trained model
+
+        Returns:
+            None
+        """
+        _, dataset_intersect_ind, checkpoint_intersect_ind = np.intersect1d(
```
rgemulla, line 105: I suggest to use ocnsistent naming (e.g., self_intersect_ind, pretrained_intersect_ind)

rgemulla: It may be more efficient to intersect the id indexes directly.

kge/model/embedder/lookup_embedder.yaml
```diff
@@ -30,6 +30,14 @@ lookup_embedder:
   initialize_args:
     +++: +++
 
+  # Initialize the model with embeddings stored in a packaged model
+  pretrain:
+      # path to packaged model containing entity/relation embeddings
```
rgemulla, line 35: Say "leave empty to not pretrain"

kge/model/embedder/lookup_embedder.yaml
```diff
@@ -30,6 +30,14 @@ lookup_embedder:
...
+      model_filename: "" 
+      # ensure that all entities/relations  can be initialized  with embeddings of 
+      #  the packaged model
```
rgemulla, line 38: Say: if false, initialikze other embeddings normally

kge/model/kge_model.py
```diff
@@ -301,6 +302,37 @@ def __init__(
...
+                    f"Initializing entities with embeddings stored in "
+                    f"{pretrained_entities_filename}"
+                )
+                packaged_checkpoint = load_checkpoint(pretrained_entities_filename)
+                packaged_model = KgeModel.create_from(packaged_checkpoint)
+                entities_ensure_all = self.get_option(
+                    "entity_embedder.pretrain.ensure_all"
+                )
+                self.init_entities_pretrained(packaged_model, entities_ensure_all)
+            if pretrained_relations_filename != "":
```
rgemulla, line 323: this is repetitive: factor out to utility method

rgemulla: The method can take a loaded checkpoint (to avoid loading twice)

kge/model/kge_model.py
```diff
@@ -426,11 +458,62 @@ def create_from(
...
+        entity_pretrain_filename = config.get_default(
+            config.get("model") + ".entity_embedder.pretrain.model_filename"
+        )
+        relation_pretrain_filename = config.get_default(
+            config.get("model") + ".relation_embedder.pretrain.model_filename"
+        )
+        config.set(config.get("model") + ".entity_embedder.pretrain.model_filename", "")
+        config.set(
+            config.get("model") + ".relation_embedder.pretrain.model_filename", ""
+        )
```
rgemulla, line 472: Why not pass a "init_for_load_only" argument to create() below and subsequently to all constructors? Then you can omit model-specific things not needed for loading without hard-coding it here.

kge/model/kge_model.py
```diff
@@ -426,11 +458,62 @@ def create_from(
...
+            ensure_all: ensure that all relations can be initialized with
+                        embeddings provided in the pre-trained model
+        """
+        self._relation_embedder.init_pretrained(
+            pretrained_model._relation_embedder._embeddings,
+            pretrained_model.dataset.relation_ids(),
+            self.dataset.relation_ids(),
+            ensure_all=ensure_all,
+        )
+
```
rgemulla, line 516: These two methods are not needed anymore once we take the ids of the embedder's dataset directly.

kge/model/embedder/lookup_embedder.py
AdrianKs: I'm not sure what you mean by intersecting on the id indexes directly

rgemulla: I mean dataset.entity_ids(), But I thought we also have inverted indexes, but apparently we don't ;). Won't matter too much anyway, since this is a one time cost.

kge/model/embedder/lookup_embedder.py
```diff
@@ -77,6 +78,49 @@ def normalize_embeddings(job):
 
             job.pre_batch_hooks.append(normalize_embeddings)
 
+    @torch.no_grad()
+    def init_pretrained(self, pretrained_embedder: "LookupEmbedder") -> None:
```
rgemulla, line 82: the __init__ method should also have init_for_load_only. For example, when this is set, none of the initialization functions needs to be called.

kge/model/embedder/lookup_embedder.py
```diff
@@ -77,6 +78,49 @@ def normalize_embeddings(job):
...
+        _, self_intersect_ind, pretrained_intersect_ind = np.intersect1d(
+            self_ids, pretrained_ids, return_indices=True
+        )
+        if self.get_option("pretrain.ensure_all") and not len(
+            self_intersect_ind
+        ) == len(self_ids):
+            raise IndexError(
+                "Not all embeddings could be initialized with the embeddings provided "
+                "in the pre-trained model"
+            )
```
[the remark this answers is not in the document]
AdrianKs: factored out to a new method in the superclass, so that we don't have to change signature/return type of the function in the superclass

kge/model/kge_model.py
```diff
@@ -291,6 +355,7 @@ def __init__(
         scorer: Union[RelationalScorer, type],
         initialize_embedders=True,
         configuration_key=None,
+        init_for_load_only=False,
```
rgemulla, line 358: initialize_embedders should probably be called create_embedders

kge/model/kge_model.py
```diff
@@ -317,8 +383,48 @@ def __init__(
...
+                pretrained_entities_model = load_pretrained_model(
+                    pretrained_entities_filename
+                )
+                if pretrained_entities_filename == pretrained_relations_filename:
+                    pretrained_relations_model = pretrained_entities_model
+                else:
+                    pretrained_relations_model = load_pretrained_model(
+                        pretrained_relations_filename
+                    )
+                if pretrained_entities_model is not None:
```
rgemulla, line 419: use the functions to get the embedders. also, make sure that subject and object embedders are identical (same object), otherwise raise an error. We can support initializing subject/object embeddings separately later on, if we actually need it

kge/model/kge_model.py
```diff
@@ -356,7 +465,9 @@ def create(
             raise Exception("Can't find {}.type in config".format(configuration_key))
 
         try:
-            model = getattr(module, class_name)(config, dataset, configuration_key)
+            model = getattr(module, class_name)(
+                config, dataset, configuration_key, init_for_load_only
```
rgemulla, line 469: pass arguments as kwargs (I think we should do this more to avoid bugs)

Issue #109, opened 2020-05-26
Change structure of sp-o and po-s indexes


AdrianKs:
The current file structure for sp-o and po-s indexes leads to pickle files, which are very large. For example for wikidata5m index-train_sp_to_o.pck more than 5GB.

If we would store the index via torch.save instead of pickle.dump this size is already reduced to 2.3GB, but loading the file and constructing the indexes still takes way to long.

We could sort the data once for sp and once for po and keep these sorted sets around. Then we just have to store the beginning and end index for each sp/po pair.

rgemulla:
We probably have to switch to torch.save anyway with PyTorch 1.5. As for the layout: yes, one could use sorted arrays + indexes, but this would mean larger changes to the code and a somewhat more cumbersome use of the indexes.

samuelbroscheit:
https://github.com/uma-pi1/kge/issues/9 is also very much related to this. Having long indexes is not necessary for our applications in the foreseeable future. Therefore we only need to convert them to long when they are use for indexing, i.e. at the very last moment.

rgemulla:
Yes, although #9 is about memory, this one about loading speed. The plan is to add a wrapper to the indexes that manage access so that we can store the index more efficiently internally.  I like the idea of addressing #9 simultaneously.

samuelbroscheit:
https://github.com/uma-pi1/kge/issues/9 also affects loading speed and general memory bandwidth bottlenecks because the storage is greatly reduced.

rgemulla:
Addressed with PR #113

PR #113, opened 2020-06-05
 Change structure of sp-o and po-s indexes

AdrianKs:
Addresses issue #109 
New index structure reduces loading time of indexes from pickle file for Wikidata5m from 681s to 21s.


rgemulla:
Do we still need `index_KvsAll_to_torch` for KvsAll training? Looks like we can now use the index more directly.

rgemulla:
Also, should we move to `torch.save` instead of pickling?

AdrianKs:
> Do we still need `index_KvsAll_to_torch` for KvsAll training? Looks like we can now use the index more directly.

I adapted the function and moved it to the index class.

AdrianKs:
> Also, should we move to `torch.save` instead of pickling?

For some reason torch.save takes up way more disk space than pickle with the new index structure. And it takes up a lot of memory while saving, which leads to an error on wikidata5m.

I propose to just not pickle the index object, since the construction is very fast.
Loading with pickle 21s vs constructing 29s on wikidata5m.
Writing the pickled index to disk takes up a lot of time and is done every time we make a small change in the dataset file or change branch.

So in total I think it is not worth writing the pickle file also given the huge file size.

rgemulla:
Sounds good. Do we have support for "selective pickling"? I think right now all indexes are pickled.

AdrianKs:
yes, unfortunately we don't have support for "selective pickling" yet

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+    def values(self):
+        values = []
+        for value in self.index.values():
+            values.append(self.data_sorted[value[0]:value[1], self.value_column])
+        return values
+
+    def items(self) -> Iterator[Tuple[Tuple[int, int], torch.Tensor]]:
+        keys = self.keys()
+        values = self.values()
+        return zip(keys, values)
```
rgemulla, line 53: Do we every use this? It's very inefficient...

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+        """
+        Construct KvsAllIndexDict
+        Args:
+            triples: data
+            key_cols: column indicators of data, giving the keys of the dictionary
+            value_column: column indicator of data, giving the values of the dictionary
+            default_factory: default return type
+        """
+        self.key_cols = key_cols
+        self.value_column = value_column
```
rgemulla, line 20: consistency: value_col

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+        self.data_sorted = self.sort_data_by_keys(triples, key_cols, value_column)
```
rgemulla, line 21: triples_sorted

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+        self.index, self.offset = self.construct_index()
```
rgemulla, line 22: offsets & why needed?

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+
+        self.default_factory = default_factory
+
+    def __getitem__(self, key, default_return_value=None):
+        try:
+            key_range = self.index[key]
```
rgemulla, line 28: start, end =

kge/indexing.py
```diff
@@ -2,6 +2,112 @@
...
+        Returns an nx2 keys tensor (rows = keys), an offset vector
+        (row = starting offset in values for corresponding key),
+        a values vector (entries correspond to values of original
+        index)
+
+        Afterwards, it holds:
+            index[keys[i]] = values[offsets[i]:offsets[i+1]]
+        """
+        keys = torch.tensor(list(self.keys()), dtype=torch.int)
+        values = self.data_sorted[:, self.value_column]
```
rgemulla, line 108: Is it actually necessary to keep all the sorted data around? Wouldn't the value column suffice?

kge/indexing.py
AdrianKs: it is indeed inefficient, but it is only used in test_dataset, since we expect working on dictionaries

kge/indexing.py
AdrianKs: Offsets used in KvsAll Training in call to index.index_tensors() (used to be index_KvsAll_to_torch) . Stored here so that we don't have to change too much in the training job.

kge/indexing.py
AdrianKs: Good point. Changed to only store value column

AdrianKs:
SInce we only keep the values around and not the complete sorted dataset the loading time of the pickle files is now reduced to 12 seconds vs 29 seconds in construction. Checked with pickle files stored on SSD not HDD. 
We should investigate the influence of torch.save vs pickle and "individual pickling" for indexes in a different PR.

PR #114, opened 2020-06-07
Update transe.py

tsafavi:
Missing a return statement in the else clause of the TransE embedding scorer.

### DIFF
--- a/kge/model/rescal.py
+++ b/kge/model/rescal.py
@@ -48,7 +48,7 @@ class RescalScorer(RelationalScorer):
                 .mm(s_emb.transpose(0, 1))
             )
         else:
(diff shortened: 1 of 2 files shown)


rufex2001:
From a quick look, isn't the "out =" what's missing?

tsafavi:
That would work too, but I was following the standards from other RelationalScorer subclasses, e.g., https://github.com/uma-pi1/kge/blob/a1a8ba04b414e58cc433257da2eddb9bd1e8ba26/kge/model/distmult.py#L24.

samuelbroscheit:
Thanks Tara! Seems a bit inconsistent in our code at the moment. Let's do "out = " to be consistent within the function.

samuelbroscheit:
Thank you!

PR #115, opened 2020-06-12
(WIP) WN11 dataset/dataset specific preprocessing

Nzteb:


.gitignore
```diff
@@ -2,9 +2,11 @@
 .idea/
 local/*
 !local/README
-/data/
+/data/*
 !data/*.sh
 !data/preprocess.py
+!data/preprocess/
```
rgemulla, line 8: move all code there; no package. I suggest download script calls preprocess/preprocess-default.py or preprocess/preprocess-<dataset>.py. All utility functions go to preprocess/util.py

.gitignore
```diff
@@ -2,9 +2,11 @@
...
+!data/__init__.py
```
rgemulla, line 9: and no __init__

data/preprocess/preprocess_util.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def process_raw_split_files(raw_split_files: dict, folder: str,  order_sop: bool = False):
+
+    if order_sop:
+        S, P, O = 0, 2, 1
+    else:
+        S, P, O = 0, 1, 2
+
+    # read data and collect entities and relations
```
rgemulla, line 15: There is still a lot of code duplication. Perhaps: move to util.py things like processing a split and updating the entity/relation dictionary as well as the yaml keys.

data/download_all.sh
```diff
@@ -201,3 +201,27 @@ else
     echo wikidata5m already prepared
 fi
 
+# wn11
+if [ ! -d "$BASEDIR/wn11" ]; then
+  echo Downloading wn11
+  cd $BASEDIR
+  # TODO this also creates a __MACOSX folder on non-mac in the zip which is deleted
+  # TODO better would be: download file from mannheim server
+  curl -O https://s3-eu-west-1.amazonaws.com/ampligraph/datasets/wordnet11.zip
```
samuelbroscheit, line 210: Yes absolutely. This was only temporary. We should put a wn11 and fb13 dataset file on our servers.

rufex2001: Done! wn11 is here: https://www.uni-mannheim.de/media/Einrichtungen/dws/pi1/kge_datasets/wn11.tar.gz
If someone provides fb13, I'll put it up as well

data/preprocess/preprocess_wn11.py
```diff
@@ -0,0 +1,159 @@
...
+    # the key of the dicts refers to the raw split
+
+    splits_positives = {
+        "train": {"file_key": "train", "file_name": "train.del"},
+        "valid": {"file_key": "valid", "file_name": "valid.del"},
+        "test": {"file_key": "test", "file_name": "test.del"},
+    }
+
+    splits_positives_wo_unseen = {
+        "train": {"file_key": "train_sample", "file_name": "train_sample.del"},
```
rgemulla, line 37: Why is this here?

data/preprocess/util.py
```diff
@@ -0,0 +1,180 @@
+import numpy as np
+
```
rgemulla, line 2: This looks good generally. Some methods lack docs (brief suffices) about what they do/their argument.s

data/preprocess/util.py
```diff
@@ -0,0 +1,180 @@
...
+
+def store_map(symbol_map: dict, filename: str):
+    with open(filename, "w") as f:
+        for symbol, index in symbol_map.items():
+            f.write(f"{index}\t{symbol}\n")
+
+
+def process_raw_split_files(raw_split_files: dict, folder: str,  order_sop: bool = False):
```
rgemulla, line 10: Perhaps: analyze_raw_splits?

rgemulla: The result could be a custom class (instead of the tuple), which needs to be passed to process_splits. This way, we can change that later on, if needed

data/preprocess/util.py
```diff
@@ -0,0 +1,180 @@
...
+     """
+    if order_sop:
+        S, P, O = 0, 2, 1
+    else:
+        S, P, O = 0, 1, 2
+
+    # the filtered split cosists of triples from the raw split where both entities
+    # and relation exist in filter_entities/filter_relation
+    if create_filtered:
+        filter_entities = kwargs["filter_entities"]
```
rgemulla, line 128: these kw arguments should be specified in the signature

data/preprocess/util.py
```diff
@@ -0,0 +1,180 @@
...
+                    filtered_pos_size += 1
+                write_triple(filtered_file_wrapper, entities, relations, t, S, P, O)
+
+        if create_filtered:
+            return pos_size, filtered_pos_size, neg_size, filtered_neg_size
+        else:
+            return pos_size, neg_size
+
+
+def write_obj_meta(config: dict):
```
rgemulla, line 166: This could be merged to a method that takes the analyzed dataset and writes entity/relation ids. All methods generally could take a (dataset) config and modify it based on what they do.

data/preprocess/util.py
```diff
@@ -0,0 +1,180 @@
...
+    for obj in ["entity", "relation"]:
+        config[f"files.{obj}_ids.filename"] = f"{obj}_ids.del"
+        config[f"files.{obj}_ids.type"] = "map"
+
+
+def write_split_meta(split_types: list, config: dict, split_sizes: dict):
```
rgemulla, line 172: E.g., this could be done as part of process split

samuelbroscheit:
I agree  with Rainers comments. Just two things to add:

1. Maybe should write a small test for those functions as well?

2. We should add an additional attribute "xvaltype" to a dataset file to define the cross validation semantics of a dataset file. This would help us, for example, in the triple classification code, to query if a split is of cross validation type "valid" or "test".

data/preprocess/preprocess_default.py
```diff
@@ -0,0 +1,119 @@
...
+from util import RawDataset
+from util import RawSplit
+from util import write_dataset_config
+from util import process_splits
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("folder", type=str)
+    parser.add_argument("--order_sop", action="store_true")
```
samuelbroscheit, line 38: Would be nice to treat this generic, this can just be a simple mapping {0:S, 1:O, 2:P}

data/preprocess/preprocess_wn11.py
```diff
@@ -0,0 +1,111 @@
...
+            "type": "triples",
+            "filename": "train.del",
+            "split_type": "train",
+        },
+        derived_sample_split_key="train_sample",
+        sample_split_options={
+            "type": "triples",
+            "filename": "train_sample.del",
+            "split_type": "train",
+        },
```
samuelbroscheit, line 39: Make this a list of DerivedSplits. In this way users can easily extend the preprocessing with a custom derived split without any hacks.

data/preprocess/util.py
```diff
@@ -0,0 +1,362 @@
...
+        """Write the derived splits of this RawSplit and collect meta-data. """
+        raise NotImplemented
+
+    def update_config(self, config: Dict) -> Dict:
+        """Update a dataset config with meta data of the derived splits."""
+        raise NotImplemented
+
+
+@dataclass
+class RawSplit(RawSplitBase):
```
samuelbroscheit, line 52: "Raw" confused me first because "raw" implies that further processing has to be done to the split, i.e. its not the final split, isn't just a "Split"?

data/preprocess/util.py
```diff
@@ -0,0 +1,362 @@
...
+    Attributes:
+        derived_split_key (str): The config key of the default split derived from
+            this raw split.
+        derived_split_options (dict): A dictionary with config options to
+            be added to the config entry of the derived default split.
+        derived_sample_split_key (str): If a key is set, an additional subsample
+            with config entry of this key will be derived from the raw split.
+        sample_split_options (dict): Dict with options to be added to the config
+            entry of the subsample split.
+        sample_size (int): Size of the subsample split.
```
samuelbroscheit, line 64: As said above: I think there should be a DerivedSplitBase class that can be implemented as FilteredDerivedSplit and SampledDerivedSplit. They should be called in a loop after the main split is finished.

samuelbroscheit (commented):
A few general comments:

- I like the structure and the new test, very nice!

 - The download_all gets quite large and also takes some time and memory for preprocessing. A nicer way would be either lazy retrieval upon use in a config (not in this PR) or only download a few (f.ex. toy, wnrr, fb237) default and the others per request on the command line. 

 - preprocess_default.py could be called something like preprocess_base.py or something to indicate the functionality and not the way it is used.

- Why are the labels for TripleClassification not stored as labels; why introduce separate splits for the positive and negative triples? Could we do both (the way you do it now and the labels)?

data/preprocess/util.py
```diff
@@ -0,0 +1,312 @@
...
+
+    def prepare_process(self, folder):
+        super().prepare_process(folder)
+        self.sample = np.random.choice(
+            len(self.parent_split.raw_data), self.sample_size, False
+        )
+
+    def process_triple(self, triple, entities, relations, **kwargs):
+        if kwargs["n"] in self.sample:
+            write_triple(
```
samuelbroscheit, line 145: call super()

data/preprocess/util.py
```diff
@@ -0,0 +1,312 @@
...
+
+    parent_split: Split = None
+    key: str = None
+    options: Dict = None
+
+    def prepare_process(self, folder):
+        self.file = open(path.join(folder, self.options["filename"]), "w")
+        self.options["size"] = 0
+
+    def process_triple(self, triple, entities, relations, **kwargs):
```
samuelbroscheit, line 83: add types for readability

samuelbroscheit (commented):
Hey, that looks very good. Just a very minor thing. I want to take the time to test it for scrutiny, but codewise I think your job is finished. Thanks!

PR #116, opened 2020-06-13
Fix embedding normalization for lookup embedder.

samuelbroscheit:
Was there because of TransE. Now using the functionality provided by Pytorch and discarding the normalization with grad.

### DIFF
--- a/kge/model/embedder/lookup_embedder.py
+++ b/kge/model/embedder/lookup_embedder.py
@@ -18,7 +18,6 @@ class LookupEmbedder(KgeEmbedder):
 
         # read config
         self.normalize_p = self.get_option("normalize.p")
-        self.normalize_with_grad = self.get_option("normalize.with_grad")
         self.regularize = self.check_option("regularize", ["", "lp"])
         self.sparse = self.get_option("sparse")
         self.config.check("train.trace_level", ["batch", "epoch"])
@@ -29,9 +28,18 @@ class LookupEmbedder(KgeEmbedder):
             self.dim = round_to_points(round_embedder_dim_to, self.dim)
 
         # setup embedder
-        self._embeddings = torch.nn.Embedding(
-            self.vocab_size, self.dim, sparse=self.sparse
-        )
+        if self.normalize_p > 0:
+            self._embeddings = torch.nn.Embedding(
+                self.vocab_size,
+                self.dim,
+                sparse=self.sparse,
+                norm_type=self.normalize_p,
+                max_norm=1.0,
+            )
+        else:
+            self._embeddings = torch.nn.Embedding(
+                self.vocab_size, self.dim, sparse=self.sparse
+            )
 
         # initialize weights
         init_ = self.get_option("initialize")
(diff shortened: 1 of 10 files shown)


rgemulla:
Please make a new PR off master with just the changes to the embedder.

kge/model/embedder/lookup_embedder.yaml
```diff
@@ -39,9 +39,6 @@ lookup_embedder:
     # l_p norm to use. Negative numbers mean do not normalize.
     p: -1.                    # common choices: 1., 2.
 
-    # Whether to push gradients through normalization.
-    with_grad: False
-
```
rgemulla, line 44: Good to remove this, but make sure the old configs still run (when this is set to false).

kge/model/embedder/lookup_embedder.py
```diff
@@ -29,9 +28,18 @@ def __init__(
...
-            self.vocab_size, self.dim, sparse=self.sparse
-        )
+        if self.normalize_p > 0:
+            self._embeddings = torch.nn.Embedding(
+                self.vocab_size,
+                self.dim,
+                sparse=self.sparse,
+                norm_type=self.normalize_p,
+                max_norm=1.0,
+            )
```
rgemulla, line 38: This is not correct. We also want vectors of norm < 1 to be normalized.

samuelbroscheit: Yes I wanted to discuss this properly this time: The only code example for embedding normalization in TransE that I found was doing it like this, i.e. only normalize when norm > 1. We were not sure why and ad hoc decided to always normalize. So the question is now what we want to do: do we replicate what somebody else did for TransE or just always normalize.

kge/model/embedder/lookup_embedder.yaml
samuelbroscheit: So throw error when "with_grad: True" and run for "with_grad: False"?

rgemulla: Yes, at the place where we process deprecated options.

kge/model/embedder/lookup_embedder.py
rgemulla: The TransE paper proposes to normalize. If we wanted max_norm (we do not for now, I think), it should be a separate option.

kge/model/embedder/lookup_embedder.yaml
samuelbroscheit: I know if you look at the changed files this partly already done in this PR.

rgemulla: Yes, once you make this PR off master it will be easy to find ;)

kge/model/embedder/lookup_embedder.py
samuelbroscheit: From Bordes' implementation seems that he always normalizes as well:

https://github.com/glorotxa/SME/blob/a1aa9fdb790fd356a17cd13d2b7b75feba2d8c82/FB15k_exp.py#L267

https://github.com/glorotxa/SME/blob/a1aa9fdb790fd356a17cd13d2b7b75feba2d8c82/model.py#L277

(I vaguely remembering looking at this code already, maybe this why we chose to always normalize back then).

PR #119, opened 2020-06-15
Fix lookup embedder normalize

samuelbroscheit:
See discussion here https://github.com/uma-pi1/kge/pull/116

### DIFF
--- a/kge/model/embedder/lookup_embedder.py
+++ b/kge/model/embedder/lookup_embedder.py
@@ -18,7 +18,6 @@ class LookupEmbedder(KgeEmbedder):
 
         # read config
         self.normalize_p = self.get_option("normalize.p")
-        self.normalize_with_grad = self.get_option("normalize.with_grad")
         self.regularize = self.check_option("regularize", ["", "lp"])
         self.sparse = self.get_option("sparse")
         self.config.check("train.trace_level", ["batch", "epoch"])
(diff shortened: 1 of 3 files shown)


samuelbroscheit:
Don't review yet.

samuelbroscheit:
Ok, now please review.

kge/model/embedder/lookup_embedder.py
```diff
@@ -61,21 +61,16 @@ def __init__(
...
+                with torch.no_grad():
+                    self._embeddings.weight.data =\
+                        torch.nn.functional.normalize(
+                            self._embeddings.weight.data, p=self.normalize_p, dim=-1
                         )
-
+            # normalize before starting a batch
             job.pre_batch_hooks.append(normalize_embeddings)
+            # normalize after an epoch (before validation and checkpoint is saved)
+            job.post_epoch_hooks.append(normalize_embeddings)
```
rgemulla, line 73: More suitable would be post_batch_hook. Otherwise looks good.

samuelbroscheit: You mean pre_batch_hooks + post_batch_hooks ?

rufex2001:
I'm guessing just normalizing after the updates, so post batch only?

kge/model/embedder/lookup_embedder.py
samuelbroscheit: @rufex2001 commented 
> I'm guessing just normalizing after the updates, so post batch only?

This would mean that the computation for the first batch after resuming or initializaton would be wrong. Or what do you think?

rufex2001:
True, those forward passes would be wrong...

kge/model/embedder/lookup_embedder.py
rgemulla: One should normalize anyway during __init__. This handles initialization and loading: the model starts with correct embeddings. Then they are renormalized whenever they are changed, i.e., after each batch.

samuelbroscheit: See https://github.com/uma-pi1/kge/pull/124 to enable this change.

rgemulla:
Superseded by #138

PR #120, opened 2020-06-15
Synchronize concurrent accesses to pickled dataset files

samuelbroscheit:
Synchronize concurrent accesses to pickled dataset files

### DIFF
--- a/kge/dataset.py
+++ b/kge/dataset.py
@@ -4,7 +4,7 @@ import csv
 import os
 import sys
 import uuid
-from multiprocessing import Lock
+from torch.multiprocessing import Lock
 
 import torch
 from torch import Tensor
--- a/kge/dataset.py
+++ b/kge/dataset.py
@@ -29,11 +29,11 @@ class Dataset(Configurable):
 
     """
 
+    # create a lock to ensure exclusive access for some operations
+    _lock = Lock()
+
     #: whether to about when an outdated cached dataset or index file is found
     _abort_when_cache_outdated = False
-
-    # create a multiprocessing lock to ensure atomic access to self._load_map
-    lock = Lock()
 
     def __init__(self, config, folder=None):
         """Constructor for internal use.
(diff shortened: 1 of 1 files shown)


kge/dataset.py
```diff
@@ -31,6 +32,9 @@ class Dataset(Configurable):
     #: whether to about when an outdated cached dataset or index file is found
     _abort_when_cache_outdated = False
 
+    # create a multiprocessing lock to ensure atomic access to self._load_map
+    lock = Lock()
```
rgemulla, line 36: _lock

kge/dataset.py
```diff
@@ -209,44 +213,45 @@ def _load_map(
...
-        else:
-            result = (dictionary, duplicates)
+        with Dataset.lock:
+            if use_pickle:
+                # check if there is a pickled, up-to-date version of the file
+                pickle_suffix = Dataset._to_valid_filename(
+                    f"-{as_list}-{delimiter}-{ignore_duplicates}.pckl"
+                )
+                pickle_filename = filename + pickle_suffix
+                result = Dataset._pickle_load_if_uptodate(None, pickle_filename, filename)
```
rgemulla, line 223: Shouldn't we rather protect _pickle_load_if_uptodate

kge/dataset.py
```diff
@@ -209,44 +213,45 @@ def _load_map(
...
+                    array[index] = value
+                result = (array, duplicates)
+            else:
+                result = (dictionary, duplicates)
 
-        if use_pickle:
-            Dataset._pickle_dump_atomic(result, pickle_filename)
-        return result
+            if use_pickle:
+                Dataset._pickle_dump_atomic(result, pickle_filename)
```
rgemulla, line 253: and _pickle_dump_atomic? Then multiple processes can still work in parallel, but loading/writing of pickles files is sequential.

AdrianKs: and we could do the writing in the background as well, since it takes quite long for large indexes and there is no need to wait for it.

rgemulla: You mean only protect the "move file" part? Sounds good to me.

AdrianKs: yes, protect the move file part and call _pickle_dump_atomic in a separate thread, so that we can compute the next index already while we write to disk.
The writing process does not need to wait for the writing to finish, since it will never read it.

rgemulla: That's too much optimzation for a bugfix PR ;). For now, I suggest to only protect move (but not the dumping) but do not start adding threading.

rgemulla:
That's not the reason for the problem (see branch use_spawn for potential fix), but still good to have. I renamed the title accordingly.

kge/dataset.py
```diff
@@ -4,6 +4,7 @@
 import os
 import sys
 import uuid
+from multiprocessing import Lock
```
rgemulla, line 7: Probably makes no difference here, but we generally use torch.multiprocessing

kge/dataset.py
```diff
@@ -215,7 +219,8 @@ def _load_map(
                 f"-{as_list}-{delimiter}-{ignore_duplicates}.pckl"
             )
             pickle_filename = filename + pickle_suffix
-            result = Dataset._pickle_load_if_uptodate(None, pickle_filename, filename)
+            with Dataset._lock:
+                result = Dataset._pickle_load_if_uptodate(None, pickle_filename, filename)
```
rgemulla, line 223: Push the lock to _pickle_load_if_uptodate

rgemulla:
One afterthought: this PR is problematic when there are multople readers (e.g., in parallel searches). Ideally, we would use a read/write lock. In the present state, I am not so sure whether this PR hurts more than it helps.

rgemulla:
(And the OS may already do the read/write locking for us anyway.)

samuelbroscheit:
_pickle_load_if_uptodate is not necessary indeed. The only minor useful thing would be if two processes arrive at _pickle_dump_atomic without checking if another one already updated the pickle. IMplemented in last commit. If not useful, then close this PR.

rgemulla:
Let's close it for now, unless we run into issues.

Issue #118, opened 2020-06-15
relational_tucker3.relation_embedder.base_embedder.dim = -1 causes error


tsafavi:
When I try to run an RT3 training job without explicitly setting `relational_tucker3.relation_embedder.base_embedder.dim`, I get a PyTorch error about creating a Tensor with a negative dimension. This is because the default config file (kge/model/relational_tucker3.yaml) sets `relational_tucker3.relation_embedder.base_embedder.dim = -1`, which doesn't appear to be updated anywhere when the embeddings are initialized. Is the -1 supposed to be referring to `relational_tucker3.relation_embedder.dim` (which is automatically set to the entity embedding dim ** 2) perhaps?  

Config file to replicate error: 
```
job.type: train
dataset.name: toy
model: relational_tucker3

relational_tucker3:
  entity_embedder:
    dim: 50
    dropout: 0.02
    regularize: 'lp'
    regularize_args:
      weight: 4e-12

  relation_embedder:
    dropout: 0.3
    regularize: 'lp'
    regularize_args:
      weight: 6e-16

train:
  max_epochs: 50
  optimizer: Adagrad
  optimizer_args:
    lr: 0.1
```

samuelbroscheit:
`relational_tucker3.relation_embedder.base_embedder.dim` has to be set explicitly 

<img src="https://raw.githubusercontent.com/ibalazevic/TuckER/master/tucker.png"/>

d_r and d_e are not depending on each other like in other KGE models.

tsafavi:
Is it worth leaving a note in the documentation that it needs to be set explicitly? Other default configs use -1 to mean that a value will get set automatically, but here that is not the case so it could be confusing for users.

rgemulla:
To clarify: 
- `relational_tucker3.relation_embedder.base_embedder.dim` (d_r) refers to the dimensions used to represent a relation and has to be specified manually. 
- `relational_tucker3.relation_embedder.dim` refers to the required embedding size for a relation. For Tucker3, this is d_e^2.
- What the RT3 embedder does by default is to project the $d_r$-sized base embedding to size $d_e^2$.

rgemulla:
Yes, it is. I'll reopen this until the docs are updated.

tsafavi:
The behavior also seems inconsistent with how entity embeddings are initialized. If I understand correctly, entity embeddings will always be initialized with a default size if not specified by the user. Why not also initialize the RT3 relation embedding dimension with a default value if not explicitly set?

[...]

PR #121, opened 2020-06-16
Update relational_tucker3.yaml

tsafavi:
Update documentation to reflect discussion in https://github.com/uma-pi1/kge/issues/118.

### DIFF
--- a/kge/model/relational_tucker3.yaml
+++ b/kge/model/relational_tucker3.yaml
@@ -7,12 +7,9 @@ relational_tucker3:
     +++: +++
   relation_embedder:
     type: tucker3_relation_embedder
-    # relation embedding size is projected to 
-    # entity_embedder.dim^2 defaults for 
-    # base_embedder.dim comes from 
-    # kge/model/embedder/tucker3_relation_embedder.yaml
-    # NOTE: base_embedder.dim must be manually set
-    # by the user to a value greater than 0
     base_embedder:
+      # Specify the desired relation embedding size here. The RT3 model projects the relation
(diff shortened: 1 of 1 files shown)


rgemulla:
Thanks!

tsafavi:
@rgemulla it's worth noting that there is still a default of base_embedder.dim = 100 in `tucker3_relation_embedder.yaml`: https://github.com/tsafavi/kge/blob/5ba82a8bea8514ed13d681b2ed9192b91164c07b/kge/model/embedder/tucker3_relation_embedder.yaml#L7

This default value is not used since it's overwritten by the -1 in `relational_tucker3.yaml`: https://github.com/tsafavi/kge/blob/5ba82a8bea8514ed13d681b2ed9192b91164c07b/kge/model/relational_tucker3.yaml#L17.

rgemulla:
Ahh, I see. I've updated the docs+configuration again and let the model use whatever default is defined in tucker3_relation_embedder.yaml (which is more consistent to what we do everywhere else).

samuelbroscheit:
Thanks @tsafavi ! Yes thats more consistent and now I understand what that issue was about :) .

samuelbroscheit:
Actually the only thing needed in relational_tucker_3.yaml is:

```
  relation_embedder:
    # See tucker3_relation_embedder.yaml for relation embedder options and
    # defaults.
    type: tucker3_relation_embedder
    +++: +++
```

rgemulla:
Yes, but this way it's easier to read/configure.

PR #122, opened 2020-06-18
Fix initialize_args in projection_embedder.py

tsafavi:
Update `projection_embedder.py` to check `initialize_args` underneath the `initialize` key first, following the convention in https://github.com/uma-pi1/kge/blob/faec621d76c1e66a4107b74a3aed14048a4327a4/kge/model/embedder/lookup_embedder.yaml#L19-L29

This allows the following example to be run:
```
job.type: train
dataset.name: toy
model: relational_tucker3

relational_tucker3:
  relation_embedder:
    initialize: normal_
    initialize_args:
      normal_:
        mean: 1.0
        std: 0.3
```


rgemulla:
Looks good. We may want to factor out the "get init" and "get args" code to a utility method at some point (e.g., the automatic setting of the uniform lower bound is not yet in the projection embedder).

PR #123, opened 2020-06-18
Factor out embedding initialization for KgeEmbedders

tsafavi:
Factor out embedding initialization to `KgeEmbedder._init_embeddings`. As per @rgemulla's comment in https://github.com/uma-pi1/kge/pull/122, two subclasses of `KgeEmbedder` rely on the same logic for initialization.


rgemulla:
Great, thanks a lot!

PR #124, opened 2020-06-21
Clean up hooks.

samuelbroscheit:
Clean up hooks. Was triggered by adressing the reviews in https://github.com/uma-pi1/kge/pull/119 to add a pre_batch_hook.

- Removed unused hooks from eval.py and entity_ranking.py that where apparently copied there from TrainJob without any purpose, their names did not make sense and post_valid_hooks was run twice apparently.

- added pre_batch_hook and pre_train_hook

- changed signature of post_train_hook, because there is no trace_entry

- move logging of num_parameters to pre_train_hook, logging it just once, instead every epoch

- guard adding a hook by querying the job.type

### DIFF
--- a/kge/job/auto_search.py
+++ b/kge/job/auto_search.py
@@ -227,14 +227,17 @@ class AutoSearchJob(SearchJob):
                 )
             )
 
-            self.trace(
-                even="search_completed",
+        job_trace = dict(
+                **job_trace,
+                event="search_completed",
                 echo=True,
                 echo_prefix="  ",
                 log=True,
                 scope="search",
                 **self.results[best_trial_index]
             )
+
+        return job_trace
 
         # DISABLED FOR NOW SINCE IDENTICAL TO BEST TRIAL
         # output parameter estimates
--- a/kge/job/auto_search.py
+++ b/kge/job/auto_search.py
@@ -91,7 +91,7 @@ class AutoSearchJob(SearchJob):
 
     # -- Main --------------------------------------------------------------------------
 
-    def _run(self, job_trace : Dict[str, Any]):
+    def _run(self, job_trace : Dict[str, Any]) -> Dict[str, Any]:
 
         # let's go
         trial_no = 0
--- a/kge/job/auto_search.py
+++ b/kge/job/auto_search.py
@@ -1,5 +1,5 @@
 import concurrent.futures
-from typing import List
+from typing import List, Any, Dict
 import torch
 from kge import Config
 from kge.config import _process_deprecated_options
@@ -91,7 +91,7 @@ class AutoSearchJob(SearchJob):
 
     # -- Main --------------------------------------------------------------------------
 
-    def _run(self):
+    def _run(self, job_trace : Dict[str, Any]):
 
         # let's go
         trial_no = 0
--- a/kge/job/entity_ranking.py
+++ b/kge/job/entity_ranking.py
@@ -400,8 +400,6 @@ class EntityRankingJob(EvaluationJob):
             event="eval_completed",
             **metrics,
         )
-        for f in self.post_epoch_trace_hooks:
-            f(self, trace_entry)
 
         # if validation metric is not present, try to compute it
         metric_name = self.config.get("valid.metric")
@@ -419,9 +417,6 @@ class EntityRankingJob(EvaluationJob):
         if was_training:
             self.model.train()
         self.config.log("Finished evaluating on " + self.eval_split + " split.")
-
-        for f in self.post_valid_hooks:
-            f(self, trace_entry)
 
         return trace_entry
 
--- a/kge/job/entity_ranking.py
+++ b/kge/job/entity_ranking.py
@@ -1,5 +1,6 @@
 import math
 import time
+from typing import Dict, Any
 
 import torch
 import kge.job
@@ -77,7 +78,7 @@ class EntityRankingJob(EvaluationJob):
         return batch, label_coords, test_label_coords
 
     @torch.no_grad()
-    def _run(self) -> dict:
+    def _run(self, job_trace : Dict[str, Any]):
 
         was_training = self.model.training
         self.model.eval()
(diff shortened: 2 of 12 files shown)


kge/job/entity_ranking.py
```diff
@@ -423,9 +421,6 @@ def merge_hist(target_hists, source_hists):
             self.model.train()
         self.config.log("Finished evaluating on " + self.eval_split + " split.")
 
-        for f in self.post_valid_hooks:
```
rgemulla, line 426: What happens with ```job.post_epoch_trace_hooks.append(append_num_parameter)``` in KgeModel?

kge/job/train.py
```diff
@@ -88,6 +88,11 @@ def __init__(
         self.num_examples = None
         self.type_str: Optional[str] = None
 
+
+        #: Hooks run after training
```
rgemulla, line 92: We is it run? Perhaps: have a pre_run and post_run hook on `Job` level instead

kge/job/train.py
```diff
@@ -109,8 +118,8 @@ def __init__(
         self.post_valid_hooks: List[Callable[[Job, Dict[str, Any]], Any]] = []
 
         #: Hooks run after training
-        #: Signature: job, trace_entry
-        self.post_train_hooks: List[Callable[[Job, Dict[str, Any]], Any]] = []
+        #: Signature: job
+        self.post_train_hooks: List[Callable[[Job], Any]] = []
```
rgemulla, line 122: We not pass in the training trace? Also, that could be a post_run_hook on Job level that gets the job and the output of the run method

kge/model/embedder/lookup_embedder.py
```diff
@@ -64,7 +64,8 @@ def normalize_embeddings(job):
                             )
                         )
 
-            job.pre_batch_hooks.append(normalize_embeddings)
+            if job.config.get("job.type") == "train":
+                job.pre_batch_hooks.append(normalize_embeddings)
```
rgemulla, line 68: WOuldn't this change output for evaluation?

rgemulla: Perhaps: also do that in the pre_run hook for eval tasks.

kge/model/kge_model.py
```diff
@@ -454,10 +454,11 @@ def prepare_job(self, job: "Job", **kwargs):
...
         self._relation_embedder.prepare_job(job, **kwargs)
 
-        def append_num_parameter(job, trace):
-            trace["num_parameters"] = sum(map(lambda p: p.numel(), self.parameters()))
+        def append_num_parameter(job):
+            self.config.log(f"num_parameters: {sum(map(lambda p: p.numel(), self.parameters()))}")
 
-        job.post_epoch_trace_hooks.append(append_num_parameter)
+        if job.config.get("job.type") == "train":
+            job.pre_train_hooks.append(append_num_parameter)
```
rgemulla, line 461: I don't like this chagne. We want models to be able to add stuff to the trace entry.

kge/job/train.py
samuelbroscheit: ok makes sense

kge/job/train.py
samuelbroscheit: but that one is from the last epoch, we can always call job.config.trace(), right?

kge/model/embedder/lookup_embedder.py
samuelbroscheit: ok, makes sense

kge/model/kge_model.py
samuelbroscheit: we can always call job.config.trace()

rgemulla: This method creates a new trace entry...

rgemulla: The use case is: have a single valid/eval entry with all the information needed

samuelbroscheit: Ahhh, of course, we can add a hook for evaluation as well. How about pre_eval_hook and post_eval_hook?

rgemulla: I think we can do this more generally using pre_run and post_run hooks, see above.

samuelbroscheit:
Ok please review. 

A few of the changes could have been added to https://github.com/uma-pi1/kge/pull/125 but this one is merged already. So I have included those changes in this PR, because some of the changes are also relevant here. 

The only thing that I haven't fixed is the use of "scope". It sometimes is used to reflect the granularity level the job  ("epoch", "batch", "example"), and then it duplicates the job type ("search", "train"). I don't know if the use of scope as in "search" and "train" is used anywhere, but I would propose to use the former meaning and also set it to "job" there, if the trace is on job level.

kge/job/auto_search.py
```diff
@@ -91,7 +91,7 @@ def get_best_parameters(self):
 
     # -- Main --------------------------------------------------------------------------
 
-    def _run(self):
+    def _run(self, job_trace : Dict[str, Any]) -> Dict[str, Any]:
```
rgemulla, line 94: Neither would I add the job_trace argument here (it can be a member of the job, if really needed) nor should the run method be required to return a Trace (at the Job level, specific jobs may do this). So, `Job.run` should have signature run() -> Any. Likewise the corresponding post_run hook. Together with the job itself, that's all the hooks will ever need. This way, we keep things simple and extensible.

kge/job/job.py
```diff
@@ -48,7 +48,6 @@ def __init__(self, config: Config, dataset: Dataset, parent_job: "Job" = None):
         self.job_id = str(uuid.uuid4())
         self.parent_job = parent_job
         self.resumed_from_job_id: Optional[str] = None
-        self.trace_entry: Dict[str, Any] = {}
```
rgemulla, line 51: Ahh, it has been like this. Keep it here.

kge/job/auto_search.py
rgemulla: TrainJob and SerachJob can make the return type more specific, of course.

rgemulla:
Superseded by #138

PR #126, opened 2020-07-01
Updated tucker3_relation_embedder.yaml

tsafavi:
With the new pretrained initialization options, the toy example for RT3 yields this error: `KeyError: 'Error accessing pretrain for key relational_tucker3.relation_embedder.pretrain.model_filename'`.  It looks like the relation embedder for Tucker needs the pretrained arguments in the config file as well.

### DIFF
--- a/kge/model/embedder/tucker3_relation_embedder.yaml
+++ b/kge/model/embedder/tucker3_relation_embedder.yaml
@@ -11,6 +11,9 @@ tucker3_relation_embedder:
   initialize: normal_          # xavier, uniform, normal
   initialize_args:
     +++: +++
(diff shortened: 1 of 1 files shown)


rgemulla:
Thanks for reporting this. To me, this sounds like a bug in the config lookup / embedder init. There shouldn't be a need to copy such config options.

@AdrianKs Can you have a look?

AdrianKs:
This was indeed a bug. Loading of pretrained embeddings is currently only supported with the LookupEmbedder but the base model still tried to initialize the ProjectionEmedder.
FIxed in 8f8ee43

PR #135, opened 2020-08-19
Revised sampling API, faster negative sampling with shared samples

rgemulla:
in batch mode. Instead of generating a tensor of all samples, directly exploits the fact that shared samples are used.

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -281,6 +281,12 @@ negative_sampling:
   # negative.
   shared: False
 
+  # if shared is True, use naive implementation of shared sampling
+  # shared samples won't be unique
+  # positive entity occurring in batch can also occur in shared negative samples
+  # faster than default shared sampling implementation
+  shared_naive: False
+
   # Whether sampling should be performed with replacement. Without replacement
   # sampling is currently only supported for shared sampling.
   with_replacement: True
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -281,8 +281,8 @@ negative_sampling:
   # negative.
   shared: False
 
-  # type of shared sampling implementation:
-
+  # Type of shared sampling implementation:
+  #
   # - naive: Each positive triples uses exactly the same set of shared samples,
   #          even if it's own positive entity is in the shared sample as a
   #          negative. Very fast.
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -281,11 +281,13 @@ negative_sampling:
   # negative.
   shared: False
 
-  # if shared is True, use naive implementation of shared sampling
-  # shared samples won't be unique
-  # positive entity occurring in batch can also occur in shared negative samples
-  # faster than default shared sampling implementation
-  shared_naive: False
+  # type of shared sampling implementation:
+  # - default : Shared samples are unique and entity of positive triple won't occur
+  #             in it's own negatives
+  # - naive   : Shared samples may not be unique, entity of positive triple might
+  #             occur in it's own negatives.
+  #             Faster than default
+  shared_type: 'default'
 
   # Whether sampling should be performed with replacement. Without replacement
   # sampling is currently only supported for shared sampling.
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -282,11 +282,15 @@ negative_sampling:
   shared: False
 
   # type of shared sampling implementation:
-  # - default : Shared samples are unique and entity of positive triple won't occur
-  #             in it's own negatives
-  # - naive   : Shared samples may not be unique, entity of positive triple might
-  #             occur in it's own negatives.
-  #             Faster than default
+
+  # - naive: Each positive triples uses exactly the same set of shared samples,
+  #          even if it's own positive entity is in the shared sample as a
+  #          negative. Very fast.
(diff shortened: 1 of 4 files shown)


kge/job/train.py
```diff
@@ -963,22 +991,50 @@ def _process_batch(self, batch_index, batch) -> TrainingJob._ProcessBatchResult:
...
+                        ]
+
+                        # repeat scores as needed for WR sampling
+                        if num_unique != num_samples:
+                            scores[:, (num_unique + 1) :] = all_scores[
+                                :, repeat_indexes[slot]
+                            ]
+
+                        # fill in positive scores
+                        scores[:, 0] = self.model.score_spo(
```
AdrianKs, line 1035: Use S, P, O instead of 0,1,2

We could also compute this once before looping over the slots. The only problem with that is, that we have to retain the graph then.

AdrianKs (commented):
We should maybe add another flag for shared sampling like shared_simple. The simple implementation does not check, if the entity itself occurs in the negatives, but shares them for all triples in the batch.
```
self._sample(positive_triples[1, :].view(1, -1), slot, num_samples).expand(len(positive_triples), num_samples)
```
This should give further time improvements.

kge/job/train.py
rgemulla: This 0 refers to the positive column, not to S. Compute once is a good idea to try!

AdrianKs: Seems like my review ended up a line above. I meant triples[:, 0], triples[: ,1], triples[:, 2] the line below

kge/config-default.yaml
```diff
@@ -281,6 +281,12 @@ negative_sampling:
   # negative.
   shared: False
 
+  # if shared is True, use naive implementation of shared sampling
+  # shared samples won't be unique
+  # positive entity occurring in batch can also occur in shared negative samples
+  # faster than default shared sampling implementation
+  shared_naive: False
```
rgemulla, line 288: I think we should instead use `shared_type=naive` to remain flexible

rgemulla: and `shared_type=default` for what we do now

AdrianKs:
I like the new structure of the sampler, but we are a little bit slower than before.
Here are results with two different fb15k configurations.

implementation  |   shared_type |   epoch_time
-----   |   -----   |   -----:
old api |   naive   | 2.8s
old api |   default | 3.3s
new api |   naive   | 2.9s
new api |   default | 3.7s


implementation  |   shared_type |   epoch_time
-----   |   -----   |   -----:
old api |   naive   | 3.4s
old api |   default | 4.1s
new api |   naive   | 3.6s
new api |   default | 4.5s

kge/util/sampler.py
```diff
@@ -188,6 +210,368 @@ def _filter_and_resample_fast(
...
+                unique_targets = None  # means all
+                column_indexes = negative_samples.contiguous().view(-1)
+            else:
+                unique_targets, column_indexes = self.unique_samples(
+                    indexes, return_inverse=True
+                )
+            self.prepare_time += time.time()
+
+            # compute all scores for slot
+            self.forward_time -= time.time()
```
AdrianKs, line 321: Should we move the following 15 lines to a little helper function to reduce duplicated code?

rgemulla:
> we are a little bit slower than before

Any idea why?

kge/util/sampler.py
rgemulla: Sure!

AdrianKs:
> > we are a little bit slower than before
> 
> Any idea why?

We are now loading the positives twice to GPU, once in the trainer and once in the sampler. But I am not sure if this already makes the difference, since the tensor is rather small..

I could not yet find another reason why. I will have another look tomorrow.

rgemulla:
> We are now loading the positives twice to GPU, once in the trainer and once in the sampler. But I am not sure if this already makes the difference, since the tensor is rather small..
Not sure. If they are already there, nothing should happen by trying to "send them again".

> I could not yet find another reason why. I will have another look tomorrow.
Sounds good!

AdrianKs:
> Not sure. If they are already there, nothing should happen by trying to "send them again".

The sampler gets the positive_triples from the collate function and therefore gets an cpu tensor. Even though process_batch gets the same cpu tensor, we move it to gpu in process_batch. This creates a new gpu tensor and therefore the sampler and the trainer and up with two different tensors.
But this huge speed issue and we could fix it by just setting ns.positive_triples = positive_triples in _process_batch.

kge/util/sampler.py
```diff
@@ -188,6 +210,368 @@ def _filter_and_resample_fast(
...
+
+        # for batch, we have a faster implementation that avoids creating the full
+        # sample tensor
+        self.prepare_time = 0.0
+        self.forward_time = 0.0
+        slot = self.slot
+        unique_targets = self._unique_samples
+        num_unique = len(unique_targets) - 1
+        triples = (
+            self.positive_triples[indexes, :] if indexes else self.positive_triples
```
AdrianKs, line 532: We are loosing the time in the following few rows indexing the chunks.
Even if we are not chunking at all, we provide the score function with an index, which just indexes all triples.
Additionally we are doing this procedure num_slots x num_chunks times, which generates an overhead

AdrianKs: yes, we get roughly the same speed now

PR #137, opened 2020-08-25
Improve Unique samples method for DefaultSharedSample

AdrianKs:
unique_samples() in DefaultSharedSample uses the default implementation of BatchNegativeSample, which is very slow.
This implementation just returns self._unique_samples if possible, drops one sample if necessary or uses the super method if we need the inverse.

Code reformatted with black.

### DIFF
--- a/kge/util/sampler.py
+++ b/kge/util/sampler.py
@@ -319,7 +319,9 @@ class BatchNegativeSample(Configurable):
 
             # compute all scores for slot
             self.forward_time -= time.time()
-            all_scores = self._score_unique_targets(model, slot, triples, unique_targets)
+            all_scores = self._score_unique_targets(
+                model, slot, triples, unique_targets
+            )
             self.forward_time += time.time()
 
             # determine indexes of relevant scores in scoring matrix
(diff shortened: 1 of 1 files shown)


kge/util/sampler.py
```diff
@@ -484,6 +486,19 @@ def __init__(
         self._drop_index = drop_index
         self._repeat_indexes = repeat_indexes
 
+    def unique_samples(self, indexes=None, return_inverse=False) -> torch.Tensor:
+        if return_inverse:
+            # slow but probably rarely used anyway
+            super(DefaultSharedNegativeSample, self).unique_samples(
+                indexes=indexes, return_inverse=return_inverse
+            )
+        elif torch.all(self._drop_index == self._drop_index[0]).item():
```
rgemulla, line 495: This needs to use indexes. Else looks good.

kge/util/sampler.py
```diff
@@ -484,6 +486,19 @@ def __init__(
...
+            # same sample dropped for every triple in the batch
+            not_drop_mask = torch.BoolTensor(len(self._unique_samples))
```
rgemulla, line 497: easier to read with torch.ones?

PR #140, opened 2020-08-28
Calculate penalty for entities only once, if subject embedder == object embedder

AdrianKs:
Calculating the penalty only once gives some small time improvements.

### DIFF
--- a/kge/model/kge_model.py
+++ b/kge/model/kge_model.py
@@ -594,14 +594,17 @@ class KgeModel(KgeBase):
             penalty_result = super().penalty(**kwargs) + self.get_p_embedder().penalty(
                 indexes=triples[:, P], **kwargs
             )
-            if self.get_s_embedder() == self.get_o_embedder():
-                entity_penalty_result = self.get_s_embedder().penalty(
-                    indexes=torch.cat(
+            if self.get_s_embedder() is self.get_o_embedder():
+                weighted = self.get_s_embedder().get_option("regularize_args.weighted")
+                entity_indexes = None
+                if weighted:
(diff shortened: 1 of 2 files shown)


kge/model/kge_model.py
```diff
@@ -591,19 +591,44 @@ def penalty(self, **kwargs) -> List[Tensor]:
...
             triples = kwargs["batch"]["triples"].to(self.config.get("job.device"))
-            return (
-                super().penalty(**kwargs)
-                + self.get_s_embedder().penalty(indexes=triples[:, S], **kwargs)
-                + self.get_p_embedder().penalty(indexes=triples[:, P], **kwargs)
-                + self.get_o_embedder().penalty(indexes=triples[:, O], **kwargs)
+            penalty_result = super().penalty(**kwargs) + self.get_p_embedder().penalty(
+                indexes=triples[:, P], **kwargs
             )
+            if self.get_s_embedder() == self.get_o_embedder():
```
rgemulla, line 597: This should be "is" instead of "==": we check for object identity instead of comparing everything.

kge/model/kge_model.py
```diff
@@ -591,19 +591,44 @@ def penalty(self, **kwargs) -> List[Tensor]:
...
+                entity_penalty_result = self.get_s_embedder().penalty(
+                    indexes=torch.cat(
```
rgemulla, line 599: These indexes are only needed in weighted, so they are unnecessarily computed for unweighted

rgemulla:
Read from my point of view.

rgemulla:
Add a changelog before merging.

PR #141, opened 2020-08-31
(WIP) compute the backward pass only once for all three slots in negative sampling

AdrianKs:
Add Flag to negative sampling to compute positive scores and backward pass only once for all three slots.

Unsure how to handle reciprocal relations in this setup. Can not provide direction, when scoring positive triples.

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -312,6 +312,13 @@ negative_sampling:
   # 'triple' (e.g., for TransE or RotatE in the current implementation).
   implementation: triple
 
+  # Whether to compute the backward pass after scoring against each slot (S, P, O)
+  # or once after scoring against all three slots.
+  # Does not work with reciprocal relations.
+  # - True: slower, but needs less memory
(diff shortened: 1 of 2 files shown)


kge/config-default.yaml
```diff
@@ -312,6 +312,13 @@ negative_sampling:
   # 'triple' (e.g., for TransE or RotatE in the current implementation).
   implementation: triple
 
+  # Whether to compute the backward pass after scoring against each slot (S, P, O)
+  # or once after scoring against all three slots.
+  # Does not work with reciprocal relations.
```
rgemulla, line 317: Move to "false".

rgemulla: Also say that if false, positive triples are scored only once (else once per slot).

rgemulla:
Another thought: shouldn't we allow this for all trainers? We may either push the backward pass into _process_batch (slow, save mem) or keep it outside (fast, more mem).

PR #143, opened 2020-08-31
Allow processing of large batches in smaller subbatches to save memory

rgemulla:
This generalizes the chunking functionality that was previously present in the negative sampling trainer to all trainers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+#### September 2020
+- PR [#143](https://github.com/uma-pi1/kge/pull/143): Allow processing of large batches in smaller subbatches to save memory
+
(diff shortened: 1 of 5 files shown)


AdrianKs (approved):
Looks good. I adapted the naming in the comments in negative sampling from chunks to subbatches.

The numbers don't change when using subbatches compared to the master branch, so everything seems to be correct.

I think we are ready to merge.

rgemulla:
Did you verify that the numbers stay for all three job types?

AdrianKs:
> Did you verify that the numbers stay for all three job types?

Yes, the numbers stay for all three train-types

rgemulla:
Thanks!

PR #147, opened 2020-09-15
Support minimization and maximization of metrics

rgemulla:
via a new config key `valid.metric_max`

@Daniel: This is largely untested & I may also have missed some places where metrics are compared.

### DIFF
--- a/kge/util/metric.py
+++ b/kge/util/metric.py
@@ -1,0 +1,36 @@
+from kge import Config
+from kge.job import Job
+from typing import Union, Iterable, List
+
+
+class Metric:
+    "Utility class for comparing metrics."
+
+    def __init__(self, metric_max: Union[Job, Config, bool]):
+        "Params: metric_max=True means higher is better."
+        if isinstance(metric_max, Job):
+            metric_max = metric_max.config
+        if isinstance(metric_max, Config):
+            metric_max = metric_max.get("valid.metric_max")
+        self._metric_max = metric_max
+
+    def better(self, metric1: float, metric2: float) -> bool:
+        if self._metric_max:
+            return metric1 > metric2
+        else:
+            return metric1 < metric2
+
+    def best(self, metrics: Iterable[float]) -> float:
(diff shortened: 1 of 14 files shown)


rufex2001:
@rgemulla the metric.py file was not added, so I can't test this yet

rgemulla:
Added.

kge/util/metric.py
```diff
@@ -0,0 +1,36 @@
...
+from typing import Union, Iterable, List
+
+
+class Metric:
+    "Utility class for comparing metrics."
+
+    def __init__(self, metric_max: Union[Job, Config, bool]):
+        "Params: metric_max=True means higher is better."
+        if isinstance(metric_max, Job):
+            metric_max = metric_max.config
```
rufex2001, line 12: Isn't the .get("valid.metric_max") missing here?

kge/util/optimizer.py
```diff
@@ -32,6 +32,24 @@ def __init__(self, config: Config, optimizer):
...
+            # check for consistency of metric-based scheduler
+            self._metric_based = name in ["ReduceLROnPlateau"]
+            if self._metric_based:
+                desired_mode = "max" if config.get("valid.metric_max") else "min"
+                if "mode" in args:
+                    if args["mode"] != desired_mode:
+                        raise ValueError(
+                            (
+                                "valid.metric_max ({}) and train.lr_scheduler_args.mode "
+                                "({}) are inconsistent."
```
rufex2001, line 44: Nice! This has to match indeed!

rufex2001 (commented):
All looks good and also works well. Tested it with a couple of runs, also tested early stopping. Just line 12 in metric.py seems off, the rest is fine!

kge/util/metric.py
rgemulla: The line right below does this.

rufex2001:
Right! I automatically read those lines as mutually exclusive...

Issue #148, opened 2020-09-21
TuckER regularization argument type error in loading checkpoint


tsafavi:
I trained and saved a TuckER model using LibKGE a few months ago. With the recent changes to `kge/model/embedder/tucker3_relation_embedder.yaml`, it looks like my old TuckER checkpoint cannot be loaded because the regularization argument `p` must be a float (e.g., 2.0) whereas it's saved as an int (e.g., 2) in my TuckER checkpoint file. 

```
Traceback (most recent call last):
  File "/hypertune-data/codex-test2/myenv/bin/kge", line 11, in <module>
    load_entry_point('libkge', 'console_scripts', 'kge')()
  File "/hypertune-data/codex-test2/kge/kge/cli.py", line 321, in main
    raise e from None
  File "/hypertune-data/codex-test2/kge/kge/cli.py", line 308, in main
    checkpoint, new_config=config, dataset=dataset
  File "/hypertune-data/codex-test2/kge/kge/job/job.py", line 120, in create_from
    checkpoint, new_config=new_config, dataset=dataset
  File "/hypertune-data/codex-test2/kge/kge/model/kge_model.py", line 560, in create_from
    config = Config.create_from(checkpoint)
  File "/hypertune-data/codex-test2/kge/kge/config.py", line 457, in create_from
    config.load_config(checkpoint["config"].clone())
  File "/hypertune-data/codex-test2/kge/kge/config.py", line 348, in load_config
    self.load_options(config.options, create, overwrite, allow_deprecated)
  File "/hypertune-data/codex-test2/kge/kge/config.py", line 342, in load_options
    self.set_all(new_options, create, overwrite)
  File "/hypertune-data/codex-test2/kge/kge/config.py", line 287, in set_all
    self.set(key, value, create, overwrite)
  File "/hypertune-data/codex-test2/kge/kge/config.py", line 224, in set
    key, type(current_value), type(value)
ValueError: key 'tucker3_relation_embedder.regularize_args.p' has incorrect type (expected <class 'float'>, found <class 'int'>)
```
What's the easiest way to fix this error?

samuelbroscheit:
Hey Tara,

The type is inferred from the default yaml configs.

I haven't checked, but if the change comes indeed from our code, then we
should have included a config rewrite to map ints to floats for
'tucker3_relation_embedder.regularize_args.p'. So a good solution would be
to add a rule to the end of kge/kge/config.py where old incompatible config
[...]

PR #149, opened 2020-09-21
Map int to float for TuckER regularize_args

tsafavi:
Map integer arguments for `tucker3_relation_embedder.regularize_args.p` to floats for backward compatibility, as per the discussion in #148.


rgemulla:
Thanks!

PR #152, opened 2020-10-06
Training Loss Evaluation

rufex2001:
This PR adds an evaluation job that is simply the training loss. Currently, support is only for computing the loss on the training split. This is part of the following PR:

https://github.com/uma-pi1/kge/pull/99

### DIFF
--- a/kge/job/eval.py
+++ b/kge/job/eval.py
@@ -148,13 +148,12 @@ class EvaluationJob(TrainingOrEvaluationJob):
 
 
 class TrainingLossEvaluationJob(EvaluationJob):
-    """ Evaluating simply by using the training loss """
+    """ Evaluating by using the training loss """
 
     def __init__(self, config: Config, dataset: Dataset, parent_job, model):
         super().__init__(config, dataset, parent_job, model)
 
         training_loss_eval_config = config.clone()
-        training_loss_eval_config.set("console.quiet", False)
         # TODO set train split to include validation data here
         #   once support is added
         #   Then reflect this change in the trace entries
@@ -167,16 +166,15 @@ class TrainingLossEvaluationJob(EvaluationJob):
             forward_only=True,
         )
 
-        # prepare training job
-        self._train_job._prepare()
-        self._train_job._is_prepared = True
-
         if self.__class__ == TrainingLossEvaluationJob:
             for f in Job.job_created_hooks:
                 f(self)
 
     def _prepare(self):
         super()._prepare()
+        # prepare training job
+        self._train_job._prepare()
+        self._train_job._is_prepared = True
 
     @torch.no_grad()
     def _evaluate(self) -> Dict[str, Any]:
@@ -185,9 +183,8 @@ class TrainingLossEvaluationJob(EvaluationJob):
 
         # create initial trace entry
         self.current_trace["epoch"] = dict(
-            type="training_loss",
+            type="training_loss_evaluation",
             scope="epoch",
-            # split=self.eval_split,
             split=self._train_job.config.get("train.split"),
             epoch=self.epoch,
         )
@@ -205,6 +202,6 @@ class TrainingLossEvaluationJob(EvaluationJob):
                 epoch_time=train_trace_entry.get("epoch_time"),
                 event="eval_completed",
                 avg_loss=train_trace_entry["avg_loss"],
-                sum_loss=train_trace_entry["sum_loss"],
+                num_examples=train_trace_entry["num_examples"],
             )
         )
--- a/kge/job/eval.py
+++ b/kge/job/eval.py
@@ -81,7 +81,6 @@ class EvaluationJob(TrainingOrEvaluationJob):
         trace_entry = self.trace(
             **self.current_trace["epoch"], echo=False, echo_prefix="  ", log=True
         )
-        self.config.log(format_trace_entry("eval_epoch", trace_entry, self.config))
         self.current_trace["epoch"] = None
 
         # reset model and return metrics
--- a/kge/job/eval.py
+++ b/kge/job/eval.py
@@ -201,7 +201,7 @@ class TrainingLossEvaluationJob(EvaluationJob):
             dict(
                 epoch_time=train_trace_entry.get("epoch_time"),
                 event="eval_completed",
-                num_examples=train_trace_entry["num_examples"],
+                size=train_trace_entry["size"],
                 avg_loss=train_trace_entry["avg_loss"],
                 avg_penalty=train_trace_entry["avg_penalty"],
                 avg_cost=train_trace_entry["avg_cost"],
--- a/kge/job/eval.py
+++ b/kge/job/eval.py
@@ -1,7 +1,8 @@
 import torch
+import time
 
 from kge import Config, Dataset
-from kge.job import Job, TrainingOrEvaluationJob
+from kge.job import Job, TrainingOrEvaluationJob, TrainingJob
 from kge.model import KgeModel
 from kge.job.trace import format_trace_entry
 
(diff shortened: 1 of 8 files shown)


kge/job/train.py
```diff
@@ -78,15 +81,22 @@ def __init__(
...
-        if self.config.get("valid.split") != "":
-            valid_conf.set("eval.split", self.config.get("valid.split"))
-        valid_conf.set("eval.trace_level", self.config.get("valid.trace_level"))
-        self.valid_job = EvaluationJob.create(
-            valid_conf, dataset, parent_job=self, model=self.model
-        )
+        self.is_forward_only = forward_only
+
+        if not self.is_forward_only:
+            self.model.train()
```
rgemulla, line 87: otherwise need to set self.model.eval()? Or is this done outside in the surrounding evaluation job?

kge/job/train.py
```diff
@@ -316,8 +344,12 @@ def run_epoch(self) -> Dict[str, Any]:
                 "split": self.train_split,
                 "batch": batch_index,
                 "batches": len(self.loader),
-                "lr": [group["lr"] for group in self.optimizer.param_groups],
+                # "lr": [group["lr"] for group in self.optimizer.param_groups],
```
rgemulla, line 347: Don't comment, remove.

kge/job/train.py
```diff
@@ -856,13 +891,30 @@ def _process_subbatch(
...
+        # But run() is never called on a training job
+        # used for a training loss eval job
+        # Run() is called on the eval job, which calls
+        # run_epoch on its training job
+        # TODO not nice but prepare methods are private
+        #   so a nicer solution requires more change
+        if not self._is_prepared:
+            super()._prepare()
+            self._prepare()
+            self._is_prepared = True
```
rgemulla, line 910: This should not be done here, but outside in TrainingEvalJob

kge/job/train.py
```diff
@@ -1002,15 +1054,32 @@ def _process_subbatch(
...
+
+        return super().run_epoch()
```
rgemulla, line 1075: This would also avoid these unnecessary repetitions.

rgemulla:
Also, the configuraiton should be updated (saying that there is now this other evaluation type).

kge/job/train.py
rufex2001: Done in the eval job

rufex2001:
@rgemulla All comments addressed in latest commit

rufex2001:
@rgemulla Please wait until I add sum_loss to tracing for merging

rufex2001:
@rgemulla I've added sum loss tracing to trainers and the training loss eval job. Two things:

1. Should we do this as simple as it is how? Or should we add this also at the batch trace level? This would mean trainers add sum_loss to their results object.
2. The solution to preparing the train job when creating it as an eval job (eval.py line 172) calls a private method from outside. Not sure this is what you meant, but this is partly why I didn't do it like this directly, but I forgot to bring it up (until now).

kge/job/eval.py
```diff
@@ -141,3 +146,66 @@ def create_from(
             new_config.set("eval.split", eval_split, create=True)
 
         return super().create_from(checkpoint, new_config, dataset, parent_job)
+
+
+class TrainingLossEvaluationJob(EvaluationJob):
+    """ Evaluating simply by using the training loss """
```
rgemulla, line 152: Drop "simply" ;)

kge/job/eval.py
```diff
@@ -141,3 +146,66 @@ def create_from(
...
+
+    def __init__(self, config: Config, dataset: Dataset, parent_job, model):
+        super().__init__(config, dataset, parent_job, model)
+
+        training_loss_eval_config = config.clone()
+        training_loss_eval_config.set("console.quiet", False)
```
rgemulla, line 158: Why is this done? Quit should be repsected, not changed.

kge/job/eval.py
```diff
@@ -141,3 +146,66 @@ def create_from(
...
+
+        self._train_job = TrainingJob.create(
+            config=training_loss_eval_config,
+            parent_job=self,
+            dataset=dataset,
+            model=model,
+            forward_only=True,
+        )
+
+        # prepare training job
```
rgemulla, line 171: Move this to _prepare of this job

kge/job/eval.py
```diff
@@ -141,3 +146,66 @@ def create_from(
...
+            split=self._train_job.config.get("train.split"),
+            epoch=self.epoch,
+        )
+
+        # run pre-epoch hooks (may modify trace)
+        for f in self.pre_epoch_hooks:
+            f(self)
+
+        # let's go
+        train_trace_entry = self._train_job.run_epoch()
```
rgemulla, line 201: metric computation is missing here. in particular, if the metric is not in the trace, the metric expr needs to be computed. the code that does this should be moved to our metric class

kge/job/train.py
```diff
@@ -490,6 +523,7 @@ def run_epoch(self) -> Dict[str, Any]:
         # add results to trace entry
         self.current_trace["epoch"].update(
             dict(
+                sum_loss=sum_loss,
```
rgemulla, line 526: I think we should drop sum_loss again. Since we already report the right avg loss (averaged over number of examples), the sum_loss can be comkputed easiliy outside (num_examples is in the trace, right?). The good thing about only having avg loss is that is makes more clear that this is our training objective (which matters to interpret regulraizitaion constants)

kge/job/eval.py
rgemulla: My bad, this is invalid!

kge/job/eval.py
rufex2001: Didn't even realize :)

kge/job/eval.py
rufex2001: Got it from the old PR, thought it was discussed already. I guess if we want to simply evaluate a model, we don't need the output from the internal training job

kge/job/eval.py
rufex2001: Right, where it should be!

kge/job/train.py
rufex2001: I Agree about clarity of training objective! Num_examples is not in the trace though, but will add it so sum loss can be computed from it

kge/job/eval.py
rufex2001: But it does raise an interesting issue: what should be added to current_trace by this type of eval job? I guess simply the training job trace? Currently, it is "harcoded" to return avg loss and sum loss.

rufex2001: I guess we should add the event eval completed and done. Then custom metrics can be defined on the train trace

rufex2001:
@rgemulla Ok, num_examples was there but called size. Renamed it to num_examples because I think that makes more sense than the notion of "epoch size". We do call it size at the batch level, which of course makes sense. If you want, I can revert this.

Also, not sure what keys to include in the trace produced by this eval job. I manually add avg_loss and num_examples from the internal training job for now. When adding everything, some things are overwritten, e.g. the event key, type key, epoch number key, etc. Also, this causes an error during the trace event, e.g. because there is already a key for job_id that comes from the already traced entry from the training job. We need to either decide which keys from the train job to include (avg cost too? avg penalty?), or which keys to exclude to prevent errors and overwritten keys. Ideally, it would be nice if all relevant keys from the train job were available to design custom metrics.

kge/job/eval.py
rgemulla: I think we do, but if we don't then quiet should be set to True, right?

kge/job/eval.py
rgemulla: good question. for now, we could trace the train entry separately (as we do for valid) and take over the loss, penalty, cost to the eval job.

rufex2001:
@rgemulla Tracing done as suggested.

kge/job/train.py
```diff
@@ -288,9 +312,12 @@ def run_epoch(self) -> Dict[str, Any]:
             epoch=self.epoch,
             split=self.train_split,
             batches=len(self.loader),
-            size=self.num_examples,
-            lr=[group["lr"] for group in self.optimizer.param_groups],
+            num_examples=self.num_examples,
```
rgemulla, line 315: Since we use size everywhere else (e.g., for the batch size), I suggest to continue using "size" here.

rgemulla (commented):
Other than this, looks good to me.

rufex2001:
@rgemulla Fixed!

rgemulla:
Alright, please add the CHANGELOG and merge

PR #154, opened 2020-10-26
store checkpoint of initialization

AdrianKs:
store a checkpoint of the initialized model at beginning of training to make experiments more reproducible.


rufex2001:
I guess this is for reproducibility across PyTorch versions? Otherwise we
can just store the seed, right?

kge/job/train.py
```diff
@@ -143,6 +143,8 @@ def _run(self) -> None:
             raise Exception(
                 f"{self.__class__.__name__} was initialized for forward only. You can only call run_epoch()"
             )
+        if self.epoch == 0:
+            self.save(self.config.checkpoint_file("init"))
```
rgemulla, line 147: Why not ```0``` instead of ```"init"```?

Also, this checkpoint file is not deleted, right? (That's good, I think.)

AdrianKs:
> I guess this is for reproducibility across PyTorch versions? Otherwise we can just store the seed, right?

This helps for reproducibility over PyTorch versions and our code changes.

kge/job/train.py
AdrianKs: Naming it 0 can lead to it being deleted.
Also this naming avoids confusion whether our epoch counter starts at 0 or 1. This way it is clear, that nothing was trained on this checkpoint

rgemulla:
@AdrianKs What's the state here? I guess only resume is open, right?

AdrianKs:
I fixed resume on checkpoint_init. I think this is ready for merge now.

kge/util/io.py
```diff
@@ -22,7 +22,10 @@ def get_checkpoint_file(config: Config, checkpoint_argument: str = "default"):
         else:
             last_epoch = config.last_checkpoint_number()
```
rgemulla, line 23: I think it would be more consistent if this would return 0 for the initial checkpoint (and the checkpoint were named _00000 as all other checkpoints). Whether or not the initial checkpoint should be retained no matter what may be a config option (default false?).

AdrianKs: renamed, but I still have to check the delete conditions, so it makes sense

kge/config.py
```diff
@@ -487,7 +487,7 @@ def last_checkpoint_number(self) -> Optional[int]:
             tried_epoch += 1
             if os.path.exists(self.checkpoint_file(tried_epoch)):
                 found_epoch = tried_epoch
-        if found_epoch > 0:
+        if found_epoch >= 0:
```
rgemulla, line 490: This doesn't work as it never even tries 0 above; the init values need to be updated as well. Also, 500 may not be enough anymore for some models...

AdrianKs: actually did work, but did not check if it actually existed. 
Improved this, so that we don't iterate over every number until 500

kge/job/train.py
```diff
@@ -224,6 +226,9 @@ def _run(self) -> None:
 
             # create checkpoint and delete old one, if necessary
             self.save(self.config.checkpoint_file(self.epoch))
+            if self.epoch == 1 and not self.config.get("train.checkpoint.keep_init"):
+                # delete init checkpoint after first epoch
+                self._delete_checkpoint(0)
```
rgemulla, line 231: I suggest to delete the init checkpoint as any other checkpoint (after keep new checkpoints) *except* if keep_init is true (then always keep it).

kge/job/train.py
```diff
@@ -238,24 +243,28 @@ def _run(self) -> None:
...
+        ):
+            self.config.log(
+                "Removing old checkpoint {}...".format(
+                    self.config.checkpoint_file(checkpoint_id)
+                )
+            )
+            os.remove(self.config.checkpoint_file(checkpoint_id))
+        else:
+            self.config.log(
+                "Could not delete old checkpoint {}, does not exits.".format(
```
rgemulla, line 263: Typo

kge/config-default.yaml
```diff
@@ -256,6 +256,9 @@ train:
     # Keep this many most recent additional checkpoints.
     keep: 3
 
+    # keep checkpoint containing initialized model for reproducibility
```
rgemulla, line 259: Rewrite to the following?

Whether to always keep checkpoint containing initialized model ('checkpoint_00000.pt') for reproducibility.

On second thoughts, I think default = true is probably a good idea.

rgemulla: Other than this, the PR is ready. If it passes all test, please go ahead and update changelog and commit.

PR #155, opened 2020-11-09
Faster read of triple files

AdrianKs:
numpy loadtxt is very slow. Using basic python operations is ~5 times faster for reading yago:

read 10x train.del  |  time
----------------    |   ---------:
old         |      31.1s
new       |      6.3s

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+#### November 2020
+- PR [#155](https://github.com/uma-pi1/kge/pull/155): Faster reading of triple files
+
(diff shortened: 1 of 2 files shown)


rgemulla:
Fine with me. I suggest, however, to try pandas.read_csv instead and use this if its fast (we already have the dependency).

Changelog is still missing.

AdrianKs:
pandas is indeed way faster:

read 10x train.del  |	time
------------  |  ------------:
numpy |	31.1s
python basics |	6.3s
pandas | 1.0s

PR #159, opened 2020-12-07
Plugin support

sfschouten:
Hi!

I've been working on support for plugins. I thought it would make sense for LibKGE to allow Models, Embedders and Datasets to be loaded from packages other than LibKGE itself. This would be especially useful if people want to implement KGE methods using LibKGE that are, for example, not relevant enough for official support within this repository to make sense.

I've implemented it by adding a 'plugins' field to the config, in which a list of packages can be specified. These packages are then included in the search for the correct configs/classes whenever an embedder/model/dataset is loaded.

Please let me know if this feature, and my way of implementing it makes sense to you.

Thanks, Stefan

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -1,3 +1,4 @@
+
 # config-default.yaml: LibKGE's default configuration options
 
 # Control output is printed to the console.
@@ -122,6 +123,14 @@ dataset:
 
   # Additional dataset specific keys can be added as needed
   +++: +++
+
+## MODULES #####################################################################
+
+# Python modules that should be searched when loading models, embedders, etc.
+# The modules are searched in order, if there are multiple components with the 
+# same name the first match is used.
+modules: [ kge.model, kge.model.embedder ]
+
 
 ## MODEL #######################################################################
 
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -1,4 +1,3 @@
-
 # config-default.yaml: LibKGE's default configuration options
 
 # Control output is printed to the console.
@@ -124,11 +123,12 @@ dataset:
   # Additional dataset specific keys can be added as needed
   +++: +++
 
+
 ## MODULES #####################################################################
 
-# Python modules that should be searched when loading models, embedders, etc.
-# The modules are searched in order, if there are multiple components with the 
-# same name the first match is used.
+# Python modules that should be searched for models, embedders, etc. LibKGE may
+# not respect the module order specified here, so make sure to use unique
+# model/embedder names.
 modules: [ kge.model, kge.model.embedder ]
 
 
--- a/kge/config.py
+++ b/kge/config.py
@@ -12,6 +12,7 @@ from enum import Enum
 import yaml
 from typing import Any, List, Dict, Optional, Union
 
+
 class Config:
     """Configuration options.
 
@@ -31,9 +32,9 @@ class Config:
             self.options = {}
 
         self.folder = folder  # main folder (config file, checkpoints, ...)
-        self.log_folder: Optional[str] = (
-            None  # None means use self.folder; used for kge.log, trace.yaml
-        )
+        self.log_folder: Optional[
+            str
+        ] = None  # None means use self.folder; used for kge.log, trace.yaml
         self.log_prefix: str = None
 
     # -- ACCESS METHODS ----------------------------------------------------------------
@@ -251,8 +252,9 @@ class Config:
         module_config.set("modules", module_names, create=True)
 
         from kge.misc import filename_in_module
-        filename = filename_in_module(self.modules(), f"{module_name}.yaml")
-        module_config.load(filename, create=True)
+
+        config_filename = filename_in_module(self.modules(), f"{module_name}.yaml")
+        module_config.load(config_filename, create=True)
 
         if "import" in module_config.options:
             del module_config.options["import"]
@@ -326,6 +328,7 @@ class Config:
             modules = modules.union(new_options.get("modules"))
             self.set("modules", list(modules), create=True)
             del new_options["modules"]
+
         # import model configurations
         if "model" in new_options:
             model = new_options.get("model")
@@ -333,6 +336,8 @@ class Config:
             # search with model as a search parameter
             if model:
                 self._import(model)
+
+        # import explicit imports
         if "import" in new_options:
             imports = new_options.get("import")
             if not isinstance(imports, list):
@@ -578,6 +583,7 @@ class Config:
 
     def modules(self) -> List[types.ModuleType]:
         import importlib
+
         return [importlib.import_module(m) for m in self.get("modules")]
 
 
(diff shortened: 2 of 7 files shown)


samuelbroscheit:
Looks very useful to me (from quickly looking at it). When the others agree as well that we should add it, then this feature should also be promoted in the README.

AdrianKs:
Yes, I think this is a useful feature. I agree that we should have a little example in the README then to show how to use it.

sfschouten:
Glad you like it! I added some info to the README.

kge/config.py
```diff
@@ -235,7 +235,7 @@ def set(
             self.log("Set {}={}".format(key, value))
         return value
 
-    def _import(self, module_name: str):
+    def _import(self, module_name: str, packages: List[str]=["kge"]):
```
rgemulla, line 238: Can we get rid of the `packages` argument? The config does (or should) have its internal value of the `plugins` option, along the lines of the `import` option.

kge/config-default.yaml
```diff
@@ -123,6 +124,13 @@ dataset:
   # Additional dataset specific keys can be added as needed
   +++: +++
 
+## PLUGINS #####################################################################
+
+# Python packages that should be searched when loading models, embedders, etc.
+# The plugin packages must use the same module hierarchy as LibKGE.
+plugins: [ ]
```
rgemulla, line 131: No need to enforce the same hierarchy, I think. All packages that contains models or embedders should be listed here.

rgemulla: `kge.model` and `kge.model.embedder` should be always included, I think.

kge/config.py
```diff
@@ -313,25 +320,31 @@ def load(
                 create=create,
                 overwrite=overwrite,
                 allow_deprecated=allow_deprecated,
+                packages=packages,
             )
 
     def load_options(
-        self, new_options, create=False, overwrite=Overwrite.Yes, allow_deprecated=True
+        self, new_options, create=False, overwrite=Overwrite.Yes, allow_deprecated=True, 
+        packages: List[str]=["kge"]
```
rgemulla, line 328: Likewise here

rgemulla (commented):
I like this idea, too; thanks for your contribution! I have only some small comments.

kge/config-default.yaml
sfschouten: Yes, I'll rephrase this. I only meant to convey that each plugin package should have at least a `model` and `model.embedder` module. Not that the entire hierarchy should be the same.

rgemulla: I was suggesting to not even require this ;) -- just specify the full module that contains the required models and/or embedders.

kge/config.py
sfschouten: Alright, I had a look at this again.
I think I've come up with a different solution that's nicer, which I will push soon.

kge/config-default.yaml
sfschouten: Ah I see, and then include kge.model and kge.model.embedder in the default value, right?

sfschouten: I suppose this would be nicest, that would also remove all occurrences of `kge` as a special case in the code.
If we do this, perhaps a better name for the config key would be `modules` instead of `plugins`?

rgemulla: Right & I agree, `modules` is more appropriate.

kge/model/kge_model.py
```diff
@@ -488,24 +489,25 @@ def create(
...
+            if hasattr(module, class_name):
+                model = getattr(module, class_name)(
+                    config=config,
+                    dataset=dataset,
+                    configuration_key=configuration_key,
+                    init_for_load_only=init_for_load_only,
+                )
+                model.to(config.get("job.device"))
+                return model
+        else:
```
rgemulla, line 507: This code is duplicated. I suggest to factor it out to kge/misc in a method along the lines of `init(modules, class_name, *args, **kwargs)`.

kge/dataset.py
```diff
@@ -96,11 +96,18 @@ def create(config: Config, preload_data: bool = True, folder: Optional[str] = No
 
         """
         name = config.get("dataset.name")
+            
+        packages = list(set(m.split(".")[0] for m in config.get("modules")))
```
rgemulla, line 100: Rename to `modules` for consistency

kge/config.py
```diff
@@ -238,22 +238,27 @@ def set(
...
+        Adds the configuration options from <module_name>.yaml to
         the configuration. Retains existing module configurations, but verifies
         that fields and their types are correct.
 
         """
-        import kge.model, kge.model.embedder
         from kge.misc import filename_in_module
+        import importlib
+        module_names = self.get_default("modules")
+        modules = [importlib.import_module(m) for m in module_names]
```
rgemulla, line 249: This is a quite frequent piece of code. Perhaps have a `get_modules()` method in `Config`?

kge/config-default.yaml
```diff
@@ -123,6 +124,13 @@ dataset:
   # Additional dataset specific keys can be added as needed
   +++: +++
 
+## MODULES #####################################################################
+
+# Python modules that should be searched when loading models, embedders, etc.
+# The plugin packages must use the same module hierarchy as LibKGE.
```
rgemulla, line 130: Drop the second sentence. Also, you may want to state that order matters here.

rgemulla (commented):
Looks great, thanks! I only made some small suggestions.

cthoyt:
I'd suggest you look into python's entrypoints, which is already part of setuptools and accomplishes a lot of this functionality

rgemulla:
Thanks for the suggestion, @cthoyt. A problem with this approach seems to be that (1) the configuration file wouldn't contain all the information anymore and (2) users need to create a package first or modify code to extend LibKGE in this way. Is this right?

cthoyt:
> Thanks for the suggestion, @cthoyt. A problem with this approach seems to be that (1) the configuration file wouldn't contain all the information anymore and (2) users need to create a package first or modify code to extend LibKGE in this way. Is this right?

I'm not sure about 1, but you're right about 2 - that's a good point.

kge/config.py
sfschouten: I did make this a method, although I only ended up calling it once because of the way dealt with the other duplicate code you pointed out.

sfschouten:
@rgemulla thanks for the great additional suggestions, I've implemented them.

Thanks @cthoyt for the suggestion. I agree with @rgemulla about the modules being listed in the config is nice. I like that it is clear from any config itself that it depends on some modules outside of LibKGE.
From my brief look at entrypoints this afternoon it does seem like it might be able to replace the searching in modules for classes/files, since plugins could 'announce' where to find them. So it might also be able to complement the current approach?
I do think that should be a separate PR though (assuming there's agreement on modules being in the config).

rgemulla:
Very nice, thanks a lot!

Issue #160, opened 2020-12-11
cannot import KgeModel, torch.fft is missing?


fratajcz:
Hi!

I just trained a model with the CLI and now i want to play with it a bit.

Unfortunately, when I import the KgeModel (from kge.model import KgeModel) I get the following Error:

```
Traceback (most recent call last):
  File "/home/fratajcz/benchmark/src/load_libkge.py", line 72, in <module>
    from kge.model import KgeModel
  File "/home/fratajcz/kge/kge/model/__init__.py", line 4, in <module>
    from kge.model.embedder.lookup_embedder import LookupEmbedder
  File "/home/fratajcz/kge/kge/model/embedder/lookup_embedder.py", line 6, in <module>
    from kge.job import Job
  File "/home/fratajcz/kge/kge/job/__init__.py", line 10, in <module>
    from kge.job.ax_search import AxSearchJob
  File "/home/fratajcz/kge/kge/job/ax_search.py", line 3, in <module>
    from ax import Models
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/ax/__init__.py", line 7, in <module>
    from ax.core import (
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/ax/core/__init__.py", line 8, in <module>
    from ax.core.arm import Arm
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/ax/core/arm.py", line 14, in <module>
    from ax.core.types import TParameterization
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/ax/core/types.py", line 11, in <module>
    from botorch.acquisition import AcquisitionFunction
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/botorch/__init__.py", line 7, in <module>
    from botorch import (
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/botorch/acquisition/__init__.py", line 7, in <module>
    from botorch.acquisition.acquisition import (
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/botorch/acquisition/acquisition.py", line 18, in <module>
    from botorch.models.model import Model
  File "/home/fratajcz/anaconda3/envs/pygeo/lib/python3.8/site-packages/botorch/models/__init__.py", line 7, in <module>
[...]

PR #163, opened 2020-12-16
Update ax version and pytorch version, adressing https://github.com/uma-pi1/kge/issues/160

samuelbroscheit:
Adress https://github.com/uma-pi1/kge/issues/160
fix  ax version and pytorch version for now
adjust to new ax API


AdrianKs:
tested on my machine, works well, will merge

at some point we need to add tests to check if everything with the search still runs smoothly

PR #164, opened 2020-12-16
Allow to easily add custom training/evaluation/search jobs

rgemulla:
This PR streamlines job creation along the lines we already do it for models, based on PR #159. It's now easy to add and use new jobs without changing LibKGE. 

See the updates in config-default.yaml as well as README.md in this PR for more info.


kge/job/train.py
```diff
@@ -120,21 +121,17 @@ def create(
...
-            )
-        elif config.get("train.type") == "1vsAll":
-            return TrainingJob1vsAll(
-                config, dataset, parent_job, model=model, forward_only=forward_only
-            )
-        else:
-            # perhaps TODO: try class with specified name -> extensibility
-            raise ValueError("train.type")
+        train_type = config.get("train.type")
+        class_name = config.get_default(f"{train_type}.class_name")
```
samuelbroscheit, line 125: Could this not also go into init_from in all occurrences, or am I missing something?

README.md
```diff
@@ -445,20 +445,30 @@ We welcome contributions to expand the list of supported models! Please see [CON
 
 ## Extending LibKGE
 
+LibKGE can be extended with new training, evaluation, or search jobs as well as
```
samuelbroscheit, line 448: Made a pass in a commit as a suggestion.

samuelbroscheit (commented):
I am wondering if this would not be the moment to finally seperate the training jobs in their own files. This has become way to large already. Then we could also put the default configs of all the jobs in their own file which would make everything more consistent and extensions more natural.

rgemulla:
I've separated out the core classes and updated the README again (note that there is no requirement on where the implementation is put). I did not yet remove the configuration of our jobs from the default config, because I think we first need to ensure that all these options are summarized somewhere with their documentation, instead of splitting it up into many files. We should eventually do this, but in the context of another PR.

README.md
```diff
@@ -453,43 +453,40 @@ KGE models implement the `KgeModel` class and generally consist of a
...
+implementation `MyClass`, you need to:
+
+1. Create a configuration file `mycomp.yaml`. You may store this file directly
+   in the LibKGE module folders (e.g., `<kge-home>/kge/model/`) or in your own
+   module folder. If you plan to contribute your code to LibKGE, we suggest to
+   directly develop in the LibKGE module folders. If you just want to play
+   around or publish your code separately from LibKGE, use your own module.
+
+2. Define all required options for your component, their default values, and
+   their types are defined in `mycomp.yaml`. We suggest to follow LibKGE's core
```
samuelbroscheit, line 468: I stumbled over this sentence "Define ... are defined". Maybe full stop after "component".

samuelbroscheit (approved):
Nice. Looks good to me, just very minor comment. What about the init_from question?

rgemulla:
> What about the init_from question?

I think `init_from` is more generic and should stay that way. That being said, these two lines could be factored out to another utility method, but I don't think it's worth it (we won't need it anymore anywhere else). 

So, ready to merge?

PR #173, opened 2021-01-18
Fix #172: Batch size inconsistency when averaging loss 

### LINKED ISSUE #172: Batch size inconsistency when averaging loss
I had a look at the different training jobs and noticed an inconsistency in how the loss is averaged between `TrainingJob1vsAll` and `TrainingJobKvsAll`/`TrainingJobNegativeSampling`. All three jobs divide the loss by `batch_size`, but in `TrainingJob1vsAll` the `batch_size` is based on an object which was indexed by `subbatch_slice` while in `TrainingJobKvsAll`/`TrainingJobNegativeSampling` the `batch_size` is based on an object which was not indexed by `subbatch_slice`.

Here are the relevant parts in the code:
```
#TrainingJob1vsAll
triples = batch["triples"][subbatch_slice].to(self.device)
batch_size = len(triples)
```
```
# TrainingJobNegativeSampling
triples = batch["triples"][subbatch_slice]
batch_negative_samples = batch["negative_samples"]
batch_size = len(batch["triples"])
```
```
#TrainingJobKvsAll
queries_subbatch = batch["queries"][subbatch_slice].to(self.device)
batch_size = len(batch["queries"])
```

I was looking at this, because I want to create unequally sized sub batches of the sub batches, based on the relation index of the triples (i.e. subsub batches). Intuitively, I would have divided the loss of a subsub batch by the size of that batch. However, the approach in `TrainingJobKvsAll`/`TrainingJobNegativeSampling` suggests I should use the overall batch size

rgemulla:


nluedema:
Looks good to me.

PR #176, opened 2021-01-19
Implemented TransH Model

Mayo42:
Hello! First of all, thank you for making this amazing project. I am using it for my research and i like it very much.

I've decided to implement the [TransH model](https://ojs.aaai.org/index.php/AAAI/article/view/8870) because I need an additional translation model besides TransE.

There is definitely room for improvement, especially when performing "sp_" and "_po" scoring, but for my needs it suffices and I want to share it.

With best regards
Phil


rgemulla:
That's great, we'd happy to extend LibKGE with your implementation!

Before a full review, I'd appreciate if you added your model to the test file [here](https://github.com/uma-pi1/kge/blob/2b693e31c4c06c71336f1c553727419fe01d4aa6/tests/test_model.py#L166) and add it to the PR. Just copy the last lines and change TransE to TransH. Then run `python -m unittest` in the LibKGE directory to verify that the various scoring functions (sp, sp_ etc) in your implementation are consistent with each other.

Mayo42 (commented):
I added TransH to unittest, but I had to add .contiguous() to the output of score_sp(s, p) because at line 53 when calling .view(-1) I would get the following error because memory was not contiguous:
`RuntimeError: view size is not compatible with input tensor's size and stride (at least one dimension spans across two contiguous subspaces). Use .reshape(...) instead.`

Test result was OK.

rgemulla:
Thanks Mayo42! I've added the soft penalties as well. Does this look OK to you? If so, I think this PR is ready to merge.

Mayo42:
I checked with the paper and found a very small mistake. I think it looks good! :)

rgemulla:
Ahh, I had actually added the "+ eps" in the denominator because it's whole purpose seems to be (the paper does not say) to avoid blow-up due to division by zero or very small numbers. It may be meant to refer to machine precision, who knows. So, I think we should keep the "+ eps" in the denominator.

rgemulla:
We should probably add a comment about this in the file itself.

Mayo42:
I see, okay this makes sense. I guess we can revert the last commit then.

Mayo42:
Okay I think we are ready to merge now. Thanks!

rgemulla:
Thanks a lot!

PR #179, opened 2021-01-28
Fix #178: Error when running negative_sampling with implementation: auto

### LINKED ISSUE #178: Error when running negative_sampling with implementation: auto
When I am running negative_sampling with implementation set to auto I get the following error:
`ValueError: Illegal value auto for key negative_sampling.implementation; allowed values are ['triple', 'batch', 'all']`

The error occurs in the init of `BatchNegativeSample`
```Python
self._implementation = self.check_option(
    "implementation", ["triple", "batch", "all"]
)
```

I had a look into the code and I think the problem is in the init of `TrainingJobNegativeSampling`
```Python
self._sampler = KgeSampler.create(config, "negative_sampling", dataset)
self._implementation = self.config.check(
    "negative_sampling.implementation", ["triple", "all", "batch", "auto"],
)
if self._implementation == "auto":
    max_nr_of_negs = max(self._sampler.num_samples)
    if self._sampler.shared:
        self._implementation = "batch"
    elif max_nr_of_negs <= 30:
        self._implementation = "triple"
    elif max_nr_of_negs > 30:
        self._implementation = "batch
```

It is checked which is the best implementation, but this decision never reaches `BatchNegativeSample`. I would propose the following fix for the init of `TrainingJobNegativeSampling`
```Python
self._sampler = KgeSampler.create(config, "negative_sampling", dataset)
self._implementation = self.config.check(
    "negative_sampling.implementation", ["triple", "all", "batch", "auto"],
)
if self._implementation == "auto":
    max_nr_of_negs = max(self._sampler.num_samples)
    if self._sampler.shared:
        self._implementation = "batch"
    elif max_nr_of_negs <= 30:
        self._implementation = "triple"
    elif max_nr_of_negs > 30:
        self._implementation = "batch
    self.config.set("negative_sampling.implementation", self._implementation)
```

nluedema:


rgemulla:
Thanks!

@rufex2001 Will the change to `auto` for the negative sampling implementation break typical TransE or RotatE configurations? If so, we should choose `triple` as a default for these models.

rufex2001:
@rgemulla negative_sampling.implementation set to auto does not work for TransE (we get the cdist error), but it does for work RotatE, at least during training. However, the published configs are all set to the deprecated spo setting for TransE, which is translated to triple, so those are fine.

rufex2001 (commented):
These changes are fine so long as we keep setting negative_sampling.implementation to triple when using TransE.

rgemulla:
I suggest to let `auto` refer to `triple` for TransE by updating the respective code in the negative sampler. Also, we should update the comment in the README to say that auto tries to pick between batch and triple heuristically. @nluedema Do you want to take a cut on this?

nluedema:
I made a change so that `auto` checks if the model uses `TransEScorer`. This should work for all the models currently implemented. However, it would not work for a model with a meta scorer that uses `TransEScorer` as its base scorer. Maybe there is a better way to check if TransE is used anywhere in the model?

Where in the README should I add a description of `auto`?

rgemulla:
Yes, documenting it in the config file as you did is good. Generally, I wouldn't write too much about how the heuristic works so that we can easily change it (perhaps write: the heuristic may change).

As for checking whether TransE is used: the cleanest way is perhaps to override the `prepare_job` method in `TransE` and let it modify `auto` to `triple`. The value of `_implementation` should then be set in `TrainingJobNegativeSampling._prepare()`.

nluedema:
@rgemulla I added a new commit with the changes, please have a look.

rgemulla:
Thanks!

Issue #183, opened 2021-02-25
Entity Ranking Imprecision caused by type casting from implicit torch.float32 to python float64


sailera19:
We noticed that there is a imprecision in calculating the entity rankings caused by type casting from the implicit torch.float32 used in torch.cumsum to the python float which is 64 bit when converting it to a list. This can be reduced to the floating point imprecision in the difference of torch.float64 and the python float by explicitly using dtype=torch.float64 in torch.cumsum.

`        hits_at_k = (
            (torch.cumsum(rank_hist[: max(self.hits_at_k_s)], dim=0, dtype=torch.float64) / n).tolist()
            if n > 0.0
            else [0.0] * max(self.hits_at_k_s)
        )
`

![image](https://user-images.githubusercontent.com/8994785/109200125-854fbc00-77a0-11eb-97e4-54bb736878cd.png)

sailera19:
Fix in #184

rgemulla:
Merged #184

PR #184, opened 2021-02-25
Fix evaluation entity ranking imprecision caused by implicit datatype cast

sailera19:
Solves issue #183


rgemulla:
Thanks, good catch!

Issue #190, opened 2021-03-06
Pretrained lookup_embedder doesn't work for reciprocal_relations_model


nluedema:
There is a problem with using pretrained relation embeddings for the lookup_embedder when using the reciprocal_relations_model. Only the first half (the non-reciprocal relation embeddings) are used, while the second half is ignored.

The problem seems to be with the function `_intersect_ids_with_pretrained_embedder` in `kge_model.py`.
```
elif "relation_embedder" in self.configuration_key:
    self_ids = self.dataset.relation_ids()
    pretrained_ids = pretrained_embedder.dataset.relation_ids()
```
The above code block only gets the first half of the relation id's and ignores the second half that was created by the reciprocal_relations_model.

rgemulla:
Thanks for reporting this bug! @AdrianKs: Will you have a look?

AdrianKs:
yes, I will have a look

AdrianKs:
fixed in #191

PR #191, opened 2021-03-08
fix loading of pretrained reciprocal relations model

AdrianKs:
fix for #190 

can not override the method _intersect_ids_with_pretrained_embedder in reciprocal_relations_model.py since the base model is created before the reciprocal relations model

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
-#### Februar 2021
+#### March 2021
+- PR [#191](https://github.com/uma-pi1/kge/pull/191): Fix loading of pretrained embeddings with reciprocal relation models
+
+#### February 2021
 - [27e8a32](https://github.com/uma-pi1/kge/commit/27e8a323d208106d7b75f4e003ea4b73c1c5d58d): improve validation time by allowing bulk KvsAll index lookup and improved history computation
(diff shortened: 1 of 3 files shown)


rgemulla:
Why is this needed? Model-specific code in a generic method is problematic from a maintainance point of view. Isn't there a clean solution (e.g., a "load embeddings" method in KgeEmbedder)?

AdrianKs:
with a "load embeddings" function in KgeEmbedder we still could not achieve a model specific initialization.

The main problem is that we would need to override the intersect_ids method inside the reciprocal relations model. But when we initialize the reciprocal relations model we first create a base model which already initializes the embedders with the method of the base model.
I am not sure how to solve this in a nice way. The design of the reciprocal relations model makes this hard.

AdrianKs:
one could make the method `_intersect_ids_with_pretrained_embedder` static and exchange the method in the init method of the reciprocal relations model
But this would also not be nice.

rgemulla:
The "right" way would be to fix the dataset or not? I.e., improve the `alt_dataset` in `ReciprocalRelationsModel` to correctly report the inverse relations with `relation_ids` by adding these relations with some suffix (e.g., `_reciprocal`).

AdrianKs:
You are right. I adapted the relation ids in the alt_dataset.

rgemulla:
Looks good (and simple)!

PR #213, opened 2021-06-11
convert key to correct type (np.int32) in kvsall index

AdrianKs:
ensure that the key in the kvsAll index always has the correct type.
fixes lookup in fast-filtering in negative sampling.


rgemulla:
Looks alright to me.

Issue #221, opened 2021-09-02
Handling of ties during evaluation.


sfschouten:
TLDR; I have been observing very high performance for a very simple model, I think a bug in the evaluation may be the cause.

I have been experimenting with learning embeddings for entities' types.
The simplest application of this is to model an entity's embedding as the mean of its types' embeddings. 
For the WN18RR dataset, this means that each entity's embedding is always that entity's type's embedding (of which there is only one for that dataset), this results in the same score being assigned to many triples.
This setting resulted in MRR scores of up to 0.723, which seems suspiciously high. And this is using the rounded_mean_rank tie handling strategy.

I have been logging a lot of values and I believe I have found the cause. In the following code snippet, floating-point issues are supposedly dealt with. 
https://github.com/uma-pi1/kge/blob/2ecac7fa8fd743f231519ad059665488440bc616/kge/job/eval_entity_ranking.py#L214-L221

However, it seems that the code is performing the opposite assignment that it should be doing. Even the comment seems to suggest the order should have been the opposite.

The way it is right now, it seems like the number of ties is pretty much always 1. This is because the value in true_scores will always match its own value, but small floating-point differences are causing it to never match the values in the chunk-based tensor of scores.
This means that ties are effectively decided 'randomly'. This does leave the question of how this 'random' tie-breaking then results in high MRR (even on the test set), but one step at a time (perhaps this tie-breaking is being 'tuned' during the hyperparameter search).

I think the code should be changed to:
```Python
                o_true_scores[o_in_chunk_mask] = scores_sp[o_in_chunk_mask, o_in_chunk]
                s_true_scores[s_in_chunk_mask] = scores_po[s_in_chunk_mask, s_in_chunk]
```
This results in the number of ties I expect for my model.

Let me know if this makes sense, more than happy to open a PR if it does.

rgemulla:
Thank you for reporting this issue. 

The main reason for the floating point issues is that (i) the true scores are computed via SPO scoring and (ii) the negative scores are computed via SP or PO scoring. This former is necessary so that the true scores are known upfront to perform evaluation in chunks, the latter is necessary to achieve reasonable efficiency.

[...]

PR #222, opened 2021-09-03
calculate true scores in evaluation using score_sp and score_po

AdrianKs:
calculate true scores in evaluation with score_sp and score_po to avoid floating point issues influencing tie breaking.
Addresses issue #221

### DIFF
--- a/kge/job/eval_entity_ranking.py
+++ b/kge/job/eval_entity_ranking.py
@@ -179,18 +179,18 @@ class EntityRankingJob(EvaluationJob):
 
             # compute true scores beforehand, since we can't get them from a chunked
             # score table
-            # here we use sp and po to stay consistent with scoring used for further
-            # evaluation. This way we avoid floating point issues influencing the
-            # tie handling.
+            # here we use score_sp and score_po to stay consistent with scoring used for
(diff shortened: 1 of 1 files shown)


AdrianKs:
@sfschouten can you please check how this change influences evaluation in your setup. If this still gives unexpected results we may have to calculate true scores with a "special chunk".

sfschouten:
I believe this fixes the problem!
Here's the before and after for a single batch https://rentry.co/rk8o9, https://rentry.co/ue86o.
As you can see the number of ties now seems a lot saner.
To make 100% sure I think I will write some tests next week, but I think this can be merged.

rgemulla:
@AdrianKs Can this be closed?

AdrianKs:
yes, this should be handled with #224 
we can reopen it if needed

AdrianKs:
Considering the issues regarding differences in the outcomes of scoring functions I reopened this PR.
We should use the same implementation for the calculation of true-scores and scores of negatives. This should allow to keep the tie-handling tolerances low.
@rgemulla do you think this is ready to merge?

rgemulla:
Fine with me. You may want to leave the old code in the source code (commented out) for clarity.

PR #223, opened 2021-09-07
entity ranking tie-handling test

sfschouten:
A new test that assigns the same score to each triple, and checks if ranks are what they should be.

Confirmed to fail on master, but succeed with #222 when `spo_different=True`.

### DIFF
--- a/tests/test_entity_ranking.py
+++ b/tests/test_entity_ranking.py
@@ -1,0 +1,116 @@
+import unittest
+import tempfile
+import yaml
+
+import torch
+from kge.job import EntityRankingJob
+from torch import Tensor
+
+from kge import Dataset
+from kge.model import KgeModel
+
+from tests.util import create_config, get_dataset_folder
+
+
+class TestModel(KgeModel):
+    """
+    Used to test proper tie handling, returns the same score for all triples.
+    Optionally returns slightly different scores depending on which scoring function is used.
+    """
+
+    SCORE = 7.291993    # arbitrary score used for all triples
+    SPO_INC = 1e-6      # arbitrary small increment for SPO triples
+
+    def __init__(self, config, dataset, spo_different):
+        """
+        Args:
+            spo_different: Whether or not SPO scores are incremented with SPO_INC. This is done to simulate a difference
+             in calculation method (e.g. for TransE SPO is calculated with F.pairwise_distance, but for SP_ / _PO
+             torch.cdist is used, which results in small differences).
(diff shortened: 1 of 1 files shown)


tests/test_entity_ranking.py
```diff
@@ -0,0 +1,116 @@
...
+from kge.job import EntityRankingJob
+from torch import Tensor
+
+from kge import Dataset
+from kge.model import KgeModel
+
+from tests.util import create_config, get_dataset_folder
+
+
+class TestModel(KgeModel):
```
rgemulla, line 15: Call this differently (it's not a test, but a dummy model).

rgemulla: The easiest way would be to use a wrapper model that takes a model and perturbs it's spo score. This then takes just a few lines.

tests/test_entity_ranking.py
```diff
@@ -0,0 +1,116 @@
...
+
+        # read trace file and check ranks are what they should be
+        trace_file = self.job.config.tracefile()
+        with open(trace_file, 'r') as file:
+            yaml_lines = file.readlines()
+
+        for yaml_line in yaml_lines:
+            yaml_line = yaml.safe_load(yaml_line)
+            if yaml_line['job_id'] == self.job.job_id \
+                    and 'event' in yaml_line.keys() and yaml_line['event'] == "example_rank":
```
rgemulla, line 103: Can use Trace.grep_entries

tests/test_entity_ranking.py
```diff
@@ -0,0 +1,116 @@
...
+class TestEntityRankingSPODifferent(BaseTestEntityRanking, unittest.TestCase):
+    def __init__(self, methodName="runTest"):
+        unittest.TestCase.__init__(self, methodName=methodName)
+        BaseTestEntityRanking.__init__(self, spo_different=True)
+
+
+class TestEntityRankingSPOSame(BaseTestEntityRanking, unittest.TestCase):
+    def __init__(self, methodName="runTest"):
+        unittest.TestCase.__init__(self, methodName=methodName)
+        BaseTestEntityRanking.__init__(self, spo_different=False)
```
rgemulla, line 116: Why not use a single `TestEntityRanking` test with two methods `test_ties_exact` and `test_ties_close` or so?

rgemulla (commented):
Looks good. Most comments are suggestions to cut code length / improve maintainability.

PR #224, opened 2021-09-08
Use torch.isclose to determine ties

sfschouten:
Addresses #221.

I think #222 should still be merged as well, the lower the tolerances can be set, the better.

Will work on tests (expanding on #223) that allow verification of good tolerance values for each model's Scorer.

### DIFF
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -492,6 +492,10 @@ entity_ranking:
   # mean_reciprocal_rank_filtered_with_test.
   filter_with_test: True
 
+  # Tie detection tolerances i.e. how close should scores be to count them as ties, see torch.isclose for more details.
+  tie_detection_atol: 1e-08
+  tie_detection_rtol: 1e-05
+
   # How to handle cases with ties between the correct answer and other answers, e.g.,
   #  Query: (s, p, ?).
   #  Answers and score: a:10, b:10, c:10, d:11, e:9
--- a/kge/config-default.yaml
+++ b/kge/config-default.yaml
@@ -492,26 +492,28 @@ entity_ranking:
   # mean_reciprocal_rank_filtered_with_test.
   filter_with_test: True
 
-  # Tie detection tolerances i.e. how close should scores be to count them as ties, see torch.isclose for more details.
-  tie_detection_atol: 1e-08
-  tie_detection_rtol: 1e-05
+  tie_handling:
 
-  # How to handle cases with ties between the correct answer and other answers, e.g.,
-  #  Query: (s, p, ?).
-  #  Answers and score: a:10, b:10, c:10, d:11, e:9
-  #  Correct: 'a'.
-  #
-  # Possible options are:
-  # - worst_rank:        Use the highest rank of all answers that have the same
-  #                      score as the correct answer. In example: 4.
-  # - best_rank:         Use the lowest rank of all answers that have the same
-  #                      score as the correct answer (competition scoring). In
-  #                      example: 2.
-  #                      DO NOT USE THIS OPTION, it leads to misleading evaluation
-  #                      results. See https://arxiv.org/pdf/1911.03903.pdf
-  # - rounded_mean_rank: Average between worst and best rank, rounded up
-  #                      (rounded fractional ranking). In example: 3.
-  tie_handling: rounded_mean_rank
+    # How to handle cases with ties between the correct answer and other answers, e.g.,
+    #  Query: (s, p, ?).
+    #  Answers and score: a:10, b:10, c:10, d:11, e:9
+    #  Correct: 'a'.
+    #
+    # Possible options are:
+    # - worst_rank:        Use the highest rank of all answers that have the same
+    #                      score as the correct answer. In example: 4.
+    # - best_rank:         Use the lowest rank of all answers that have the same
+    #                      score as the correct answer (competition scoring). In
+    #                      example: 2.
+    #                      DO NOT USE THIS OPTION, it leads to misleading evaluation
+    #                      results. See https://arxiv.org/pdf/1911.03903.pdf
+    # - rounded_mean_rank: Average between worst and best rank, rounded up
+    #                      (rounded fractional ranking). In example: 3.
+    type: rounded_mean_rank
+
+    # Tie detection tolerances i.e. how close should scores be to count them as ties, see torch.isclose for more details.
+    atol: 1e-08
+    rtol: 1e-05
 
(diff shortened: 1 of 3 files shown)


kge/config-default.yaml
```diff
@@ -492,6 +492,10 @@ entity_ranking:
   # mean_reciprocal_rank_filtered_with_test.
   filter_with_test: True
 
+  # Tie detection tolerances i.e. how close should scores be to count them as ties, see torch.isclose for more details.
+  tie_detection_atol: 1e-08
+  tie_detection_rtol: 1e-05
+
```
rgemulla, line 498: I suggest to group as `tie_handling.type`, `tie_handling.atol` and `tie_handling.rtol` + for backwards compatiblity a rewrite of `tie_handling` (when str) to `tie_handling.type` (see bottom of `config.py`)

rgemulla (commented):
Thanks, looks good to me!

@AdrianKs Any other thoughts?

kge/config.py
```diff
@@ -746,6 +746,9 @@ def rename_value_re(key_regex, old_value, new_value):
                     renamed_keys.add(key)
         return renamed_keys
 
+    # 08.09.21
+    rename_key("entity_ranking.tie_handling", "entity_ranking.tie_handling.type")
```
rgemulla, line 750: Does this work if one uses a config that already uses `entity_ranking.tie_handling.type`?

AdrianKs: I think this should work, at least we also did it for the optimizer like that:
```python
    # 09.10.20
    rename_key("train.optimizer", "train.optimizer.default.type")
    rename_keys_re("^train\.optimizer_args", "train.optimizer.default.args")
```

AdrianKs:
looks good to me.
I aggree that we should still merge #222 as it also avoids bugs for people implementing spo, po and sp options that don't give the same results

rgemulla:
@AdrianKs If fine, please go ahead and merge this PR.

As for #222, I am not so sure. spo vs sp_ vs _po scoring is already covered in our tests (testing closeness, in fact) and the change does make evaluation somewhat more involved. A better approach is to assume closeness, but not exact equality, during development.

sfschouten:
I think it would be a good idea to add a check like the following: 
https://github.com/sfschouten/kge/blob/34730e138a0dceff864603dbab0d5465403f2b8a/kge/job/eval_entity_ranking.py#L223-L231
This stops poorly chose tolerance values from being silently messing up the evaluation. I think this is valuable because from my testing whether or not tolerances are wide enough can actually depend on other hyperparameters.

Not totally sure if raising an Error is the best thing to do upon detection.
What do you think @rgemulla @AdrianKs ?

rgemulla:
I like this idea very much! Are there any other sensible options than to raise an error?

sfschouten:
I suppose not actually, I was thinking about a warning, but probably doesn't ever make sense to continue evaluating once this happens.

AdrianKs:
If we are not merging #222 , should we add a hint to this error message, that the spo, sp_ and _po implementations of the used model may not return the same values?
This is important for users implementing new models without extending the unit tests for the new model.

rgemulla:
Yes, also a good idea. One may even point out the (spo) triple in question.

Is the trained model at this time already checkpointed? Otherwise, one problem of an error may be that the training epoch might get lost.

sfschouten:
> Is the trained model at this time already checkpointed? Otherwise, one problem of an error may be that the training epoch might get lost.

I think that the checkpoint does get lost with an error. However, this might make sense because since we cannot evaluate, we also do not know whether this checkpoint is better than the previous best. Maybe resuming from the last checkpoint that was evaluated correctly is okay for now?

AdrianKs:
overall it would be better though if we checkpointed the model before evaluation. Otherwise, if someone hits an OOM with evaluation the complete epoch will be lost. 
But, I think this can be a sperate PR.

rgemulla:
Yes, can be a separate PR. It may be a good idea to create a feature-request issue for this, just to keep this in mind.

kge/config-default.yaml
```diff
@@ -492,22 +492,28 @@ entity_ranking:
...
+    # - best_rank:         Use the lowest rank of all answers that have the same
+    #                      score as the correct answer (competition scoring). In
+    #                      example: 2.
+    #                      DO NOT USE THIS OPTION, it leads to misleading evaluation
+    #                      results. See https://arxiv.org/pdf/1911.03903.pdf
+    # - rounded_mean_rank: Average between worst and best rank, rounded up
+    #                      (rounded fractional ranking). In example: 3.
+    type: rounded_mean_rank
+
+    # Tie detection tolerances i.e. how close should scores be to count them as ties, see torch.isclose for more details.
```
rgemulla, line 514: Apply formatting. I suggest to write a bit more here to be clear. E.g., "# Tie detection tolerances. Two scores count as a tie whenever they fall within this tolerance; see torch.isclose for details. Important to increase numerical stability (see <link-to-bug-report>."

kge/job/eval_entity_ranking.py
```diff
@@ -217,8 +219,22 @@ def _evaluate(self):
...
+                                             rtol=self.tie_rtol, atol=self.tie_atol)
+                close_check &= torch.allclose(scores_po[s_in_chunk_mask, s_in_chunk], s_true_scores[s_in_chunk_mask],
+                                              rtol=self.tie_rtol, atol=self.tie_atol)
+                if not close_check:
+                    diff_a = torch.abs(scores_sp[o_in_chunk_mask, o_in_chunk] - o_true_scores[o_in_chunk_mask])
+                    diff_b = torch.abs(scores_po[s_in_chunk_mask, s_in_chunk] - s_true_scores[s_in_chunk_mask])
+                    diff_all = torch.cat((diff_a, diff_b))
+                    self.config.log(f"Tie-handling: mean difference between scores was: {diff_all.mean()}.")
+                    self.config.log(f"Tie-handling: max difference between scores was: {diff_all.max()}.")
+                    raise ValueError("Tie-handling tolerances set too low! The same triples scored twice did not fall "
```
rgemulla, line 235: It's not necessarily an error in the tolerance, could also be the implementation. Perhaps write: "Error in tie handling. A triple was assigned different scores by the spo/sp_/_po scoring implementations during evaluation. Verify model implementation or consider increasing tie-handling tolerances."

rgemulla (commented):
Looks good to me. Only two suggested textual changes, then can be merged.

Thanks @sfschouten

PR #225, opened 2021-09-13
deepcopy _meta dict in reciprocal relations

sfschouten:
Quick fix for a small bug. 
Only the reference to the `_meta` dictionary in each dataset was being copied, so the number of relations was doubling each trial while performing a search.

### DIFF
--- a/kge/model/reciprocal_relations_model.py
+++ b/kge/model/reciprocal_relations_model.py
@@ -26,10 +26,11 @@ class ReciprocalRelationsModel(KgeModel):
         # Using a dataset with twice the number of relations to initialize base model
         alt_dataset = dataset.shallow_copy()
         alt_dataset._num_relations = dataset.num_relations() * 2
-        reciprocal_relation_ids = [
-            rel_id + "_reciprocal" for rel_id in alt_dataset.relation_ids()
-        ]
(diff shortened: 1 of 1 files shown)


kge/model/reciprocal_relations_model.py
```diff
@@ -26,10 +27,10 @@ def __init__(
...
         alt_dataset = dataset.shallow_copy()
         alt_dataset._num_relations = dataset.num_relations() * 2
-        reciprocal_relation_ids = [
-            rel_id + "_reciprocal" for rel_id in alt_dataset.relation_ids()
-        ]
-        alt_dataset._meta["relation_ids"].extend(reciprocal_relation_ids)
+        alt_dataset._meta = copy.deepcopy(dataset._meta)
+        alt_dataset._meta["relation_ids"].extend([
+            rel_id + "_reciprocal" for rel_id in dataset.relation_ids()
+        ])
```
rgemulla, line 33: Good catch! Copying everything is too heavy though (meta data may actually be quite substantial). What we want is shallow copy of the dict and a deep copy of just the relation_ids.

sfschouten: Right, that makes sense.

rgemulla:
Thanks!

PR #249, opened 2021-11-25
overwrite dataset-config options only if default is currently set

AdrianKs:
Config options defined in config.yaml or directly in the terminal regarding the dataset are currently overwritten by dataset.yaml; e.g. filenames.
We print to the console the options defined by the user but do not apply them.

This PR fixes this behavior and dataset.yaml only overwrites default options, now.

### DIFF
--- a/kge/cli.py
+++ b/kge/cli.py
@@ -251,11 +251,6 @@ def main():
         # disable processing of outdated cached dataset files globally
         Dataset._abort_when_cache_outdated = args.abort_when_cache_outdated
 
-        # log configuration
-        config.log("Configuration:")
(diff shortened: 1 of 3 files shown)


rgemulla:
Ok with me. A few comments:

1. Perhaps add a comment to the default_options field stating what it's used for.
2. Is the config printed the one that's actually used?
3. If a config option is not applied because it's default has been changed, we may want to print a warning.

AdrianKs:
Fixed the nested lookup for the default options.

1. Done
2. Moved dumping of config after dataset and job creation.
3. Done

PR #261, opened 2022-04-13
updated the util code to buffer mode write

bhanu77prakash:
As mentioned in the latest issue, I have tried to convert the write mode from line by line to a buffer mode write. I tried running the updated script on the toy dataset and the outputs are exactly same.

### DIFF
--- a/data/preprocess/util.py
+++ b/data/preprocess/util.py
@@ -8,6 +8,7 @@ import os
 
 ### DEFAULT PREPROCESS UTILS ##########################################################
 
+BUFFER_SIZE = 1000000
 
 @dataclass
 class RawSplit:
@@ -52,6 +53,7 @@ class RawSplit:
             for split in self.splits:
                 split.process_triple(t, entities, relations, n=n)
         for split in self.splits:
+            dump_buffer_to_file(split.file_buffer, split.file)
             split.file.close()
 
     def update_config(self, config: Dict) -> Dict:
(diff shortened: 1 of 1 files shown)


data/preprocess/util.py
```diff
@@ -252,10 +256,18 @@ def store_map(symbol_map: Dict, filename: str):
...
 
+def dump_buffer_to_file(buffer, fptr):
+    """Dump buffer to file."""
+    fptr.write(f"{buffer}")
 
-def write_triple(f, ent, rel, t, S, P, O):
+def write_triple(f, buffer, ent, rel, t, S, P, O):
     """Write a triple to a file. """
-    f.write(str(ent[t[S]]) + "\t" + str(rel[t[P]]) + "\t" + str(ent[t[O]]) + "\n")
+    buffer += f"{str(ent[t[S]])}\t{str(rel[t[P]])}\t{str(ent[t[O]])}\n"
```
AdrianKs, line 265: I think it is faster, to make the buffer a list of strings and finally join this list to a single string
```python
buffer_list = []
buffer_list.append(new_triple_string)

buffer = ""
buffer.join(buffer_list)
```

data/preprocess/util.py
```diff
@@ -252,10 +256,18 @@ def store_map(symbol_map: Dict, filename: str):
...
+    if len(buffer) > BUFFER_SIZE:
+        dump_buffer_to_file(buffer, f)
```
AdrianKs, line 267: did you make sure, that the rest of the buffer is written to file at every necessary position? you did so in the RawSplit but process_triple is called also in other splits

AdrianKs:
Thanks, this is a good idea. Some small changes are needed though.

PR #265, opened 2022-08-03
N3 implementation for ComplEx and Reciprocal ComplEx

Kenkoko:
Hi @rgemulla @rufex2001,

This is the implementation of N3 - Chen et al. (2021). 

Let $u$ be an embedding vector. The formula for L3 is: $$\|u \|^3_p = \sum_{j=1}^{n_d}|u_j|^3$$

The main difference between N3 and L3  implementation is that Chen et al. (2021) considered an embedding vector $u$ as a complex vector while LibKGE considered an embedding vector as a real vector $u$. 

If $u$ is a complex vector, $u_j = u'_j + iu''_j$  where $u'_j$ is the real part and $u''_j$ is the imaginary. The L3 would be:

$$\|u \|^3_p = \sum_{j=1}^{n_d}|u_j|^3 = \sum_{j=1}^{n_d}\left(\sqrt{{u'}_j^2 + {u''}_j^2}\right)^3$$

If $u$ is a real vector, the L3 would be:

$$\|u \|^3_p = \sum_{j=1}^{n_d}|u_j|^3 $$

where $|u_j|$ is absolute value of $u_j$.

This pull request included a new function that calculates the complex vectors' norm before cubing. The calculation is only conducted for the ComplEx and Reciprocal ComplEx models.

Please let me know what you think about the code. If you have any suggestions, please feel free to suggest.

### DIFF
--- a/kge/model/complex.py
+++ b/kge/model/complex.py
@@ -1,7 +1,8 @@
 import torch
 from kge import Config, Dataset
 from kge.model.kge_model import RelationalScorer, KgeModel
-
+from typing import List
+from torch import Tensor
 
 class ComplExScorer(RelationalScorer):
     r"""Implementation of the ComplEx KGE scorer.
@@ -60,3 +61,6 @@ class ComplEx(KgeModel):
             configuration_key=configuration_key,
             init_for_load_only=init_for_load_only,
         )
+    
+    def penalty(self, **kwargs) -> List[Tensor]:
+        return super().penalty(is_complex= True, **kwargs)
--- a/kge/model/complex.py
+++ b/kge/model/complex.py
@@ -61,6 +61,3 @@ class ComplEx(KgeModel):
             configuration_key=configuration_key,
             init_for_load_only=init_for_load_only,
         )
-    
-    def penalty(self, **kwargs) -> List[Tensor]:
-        return super().penalty(is_complex= True, **kwargs)
--- a/kge/model/complex.py
+++ b/kge/model/complex.py
@@ -2,6 +2,7 @@ import torch
 from kge import Config, Dataset
 from kge.model.kge_model import RelationalScorer, KgeModel
 
+
 class ComplExScorer(RelationalScorer):
     r"""Implementation of the ComplEx KGE scorer.
 
--- a/kge/model/complex.py
+++ b/kge/model/complex.py
@@ -1,8 +1,6 @@
 import torch
 from kge import Config, Dataset
 from kge.model.kge_model import RelationalScorer, KgeModel
-from typing import List
-from torch import Tensor
 
 class ComplExScorer(RelationalScorer):
     r"""Implementation of the ComplEx KGE scorer.
--- a/kge/model/complex.yaml
+++ b/kge/model/complex.yaml
@@ -4,7 +4,9 @@ complex:
   class_name: ComplEx
   entity_embedder:
     type: lookup_embedder
+    vector_space: complex
     +++: +++
   relation_embedder:
     type: lookup_embedder
+    vector_space: complex
     +++: +++
--- a/kge/model/complex.yaml
+++ b/kge/model/complex.yaml
@@ -4,9 +4,9 @@ complex:
   class_name: ComplEx
   entity_embedder:
     type: lookup_embedder
-    vector_space: complex
+    space: complex
     +++: +++
   relation_embedder:
     type: lookup_embedder
-    vector_space: complex
+    space: complex
     +++: +++
--- a/kge/model/embedder/lookup_embedder.py
+++ b/kge/model/embedder/lookup_embedder.py
@@ -25,7 +25,7 @@ class LookupEmbedder(KgeEmbedder):
 
         # read config
         self.normalize_p = self.get_option("normalize.p")
-        self.regularize = self.check_option("regularize", ["", "lp"])
+        self.regularize = self.check_option("regularize", ["", "lp", "n3"])
         self.sparse = self.get_option("sparse")
         self.config.check("train.trace_level", ["batch", "epoch"])
         self.vocab_size = vocab_size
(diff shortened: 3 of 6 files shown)


kge/model/complex.py
```diff
@@ -60,3 +61,6 @@ def __init__(
             configuration_key=configuration_key,
             init_for_load_only=init_for_load_only,
         )
+    
+    def penalty(self, **kwargs) -> List[Tensor]:
+        return super().penalty(is_complex= True, **kwargs)
```
rgemulla, line 66: ComplEx isn't the only model that uses complex embeddings. Also, I suggest to move the is_complex flag to lookup embedder instead.

kge/model/embedder/lookup_embedder.py
```diff
@@ -107,21 +107,31 @@ def _embeddings_all(self) -> Tensor:
     def _get_regularize_weight(self) -> Tensor:
         return self.get_option("regularize_weight")
 
+    def normalize_complex(self, parameters) -> Tensor:
```
rgemulla, line 110: This is not normalizing, but computing the element-wise absolute value. I suggest to rename to _abs_complex.

kge/model/embedder/lookup_embedder.yaml
```diff
@@ -50,7 +50,9 @@ lookup_embedder:
     p: -1.                    # common choices: 1., 2.
 
   # Whether and how embeddings should be regularized. Possible values are '' (do
-  # not regularize) or lp (defaults to p=2; else set regularize_args.p)
+  # not regularize) or lp (defaults to p=2; else set regularize_args.p) or 
+  # n3 (which is l3 but the embedding vectors are treated as complex vector when
```
rgemulla, line 54: which is l3 applied to the element-wise absolute value of complex embeddings)

rgemulla (commented):
Thanks! I've made a few comment.

kge/model/complex.py
Kenkoko: Hi Prof Gemulla, 

Instead of using is_complex, I created a new configuration in the YAML file, which is vector_space (accepts real and complex values). The computing N3 will base on the value of vector_space configuration to compute the penalty for embedding vectors.

Please let me know what you think about this change. 

For the other model - RotatE, I will change it when we finalize the final approach.

kge/model/embedder/lookup_embedder.py
Kenkoko: changed ^^ thank for suggestion.

kge/model/embedder/lookup_embedder.yaml
Kenkoko: Also fixed ^^.

Kenkoko:
Hi @rgemulla,

I have made some changes; please let me know what you think about those changes.

kge/model/complex.yaml
```diff
@@ -4,7 +4,9 @@ complex:
   class_name: ComplEx
   entity_embedder:
     type: lookup_embedder
+    vector_space: complex
```
rgemulla, line 7: Rename to "space", default "euclidean".

kge/model/embedder/lookup_embedder.py
```diff
@@ -25,10 +25,11 @@ def __init__(
 
         # read config
         self.normalize_p = self.get_option("normalize.p")
-        self.regularize = self.check_option("regularize", ["", "lp"])
+        self.regularize = self.check_option("regularize", ["", "lp", "n3"])
         self.sparse = self.get_option("sparse")
         self.config.check("train.trace_level", ["batch", "epoch"])
         self.vocab_size = vocab_size
+        self.vector_space = self.check_option("vector_space", ["real", "complex"])
```
rgemulla, line 32: check that n3 is only enabled when space is marked "complex"

kge/model/embedder/lookup_embedder.py
```diff
@@ -134,6 +145,10 @@ def penalty(self, **kwargs) -> List[Tensor]:
                     kwargs["indexes"], return_counts=True
                 )
                 parameters = self._embeddings(unique_indexes)
+
+                if self.regularize == "n3" and self.vector_space == 'complex':
+                    parameters = self._abs_complex(parameters)
+
                 if p % 2 == 1:
                     parameters = torch.abs(parameters)
```
rgemulla, line 153: This abs is not needed for n3

rgemulla (commented):
Looks great! I've added only a few very minor comments.

kge/model/complex.yaml
Kenkoko: changed ^^

kge/model/embedder/lookup_embedder.py
Kenkoko: added more condition ^^

kge/model/embedder/lookup_embedder.py
Kenkoko: Please let me know if my implementation is exactly what you are looking for or not ^^

If it's exactly what you are looking for, should I raise an error when a user uses N3 in euclidian space?

Kenkoko:
Hi @rgemulla,

I have made some changes based on your comments; please let me know what you think.

kge/model/embedder/lookup_embedder.py
rgemulla: Yes, but this should be the case already with your implementaiton.

rgemulla:
Looks great!

@rufex2001 If ok with you as well, please update CHANGELOG and merge.

Kenkoko:
Hi @rgemulla, @rufex2001,

Before merging, I would like to reproduce the results from Chen et al. (2021) first to make sure that everything works as we expect.

Kenkoko:
Hi @rgemulla, @rufex2001,

I have successfully reproduced the Chen et al. (2021) result. 

```
2022-09-01 05:08:03.621303 [56e0d1a2]   mean_reciprocal_rank_filtered_with_test: 0.37134475468705447
2022-09-01 05:08:03.621313 [56e0d1a2]   mean_reciprocal_rank_filtered_with_test_1-1: 0.5777754184014783
2022-09-01 05:08:03.621329 [56e0d1a2]   mean_reciprocal_rank_filtered_with_test_1-N: 0.29677387943432404
2022-09-01 05:08:03.621339 [56e0d1a2]   mean_reciprocal_rank_filtered_with_test_M-1: 0.4563853876377509
2022-09-01 05:08:03.621349 [56e0d1a2]   mean_reciprocal_rank_filtered_with_test_M-N: 0.34782264950258296
2022-09-01 05:08:03.621359 [56e0d1a2]   metric_name: mean_reciprocal_rank_filtered_with_test
```
 
I would like to send you the log file for detail discussion. 

[kge.log](https://github.com/uma-pi1/kge/files/9469399/kge.log)

rgemulla:
@Kenkoko Thanks! Can you also post the config file you used here?

@rufex2001 If you agree, please update changelog and merge.

Kenkoko:
Hi @rgemulla, @rufex2001 

GitHub doesn't allow me to upload a Yaml file. So I changed the config.yaml file to txt file.

Please let me know if you need anything further ^^

[config.txt](https://github.com/uma-pi1/kge/files/9742370/config.txt)

rufex2001:
Looks good! @Kenkoko thanks for the contribution and the config file for reproduction!

PR #271, opened 2023-02-16
fix incorrect dependency version requirement

ppoffice:
See https://github.com/pypa/packaging/issues/645

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     author_email="rgemulla@uni-mannheim.de",
     packages=["kge"],
     install_requires=[
-        "numpy>=1.19.*",
+        "numpy==1.19.*",
         "torch>=1.7.1",
         "pyyaml",
         "pandas",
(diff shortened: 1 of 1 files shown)


AdrianKs:
Thank you for the PR. 
We would like to be able to always use the newest version of the packages if possible. Therefore, it would be preferable to remove the wildcard instead of the `>=`. 
`numpy>=1.19.0`

ppoffice:
@AdrianKs Here you go.

BTW, I think using `compatible release` versions instead of the latest version will ensure dependency compatibility. For instance, you can't prevent pip from installing `numba 1.0.0` when it's available if you set `numba>=0.50.0` as a dependency. And usually a major upgrade like `0.x.x` to `1.0.0` introduces breaking changes.

ppoffice:
Would you kindly merge this PR if there is no more issue with it? Thanks.

AdrianKs:
Yes, sure. Sorry for the late reply.
Thank you for the pointer to using `~=`.
We will merge as it is for now with `>=` and discuss later on how we want to treat major releases of packages.

